PR #21, opened 2022-05-21
null vector cause crash

Dumbstep:
image
image


BenLubar:
I don't think uninitialized floats can cause this error, but this change won't hurt.

PR #42, opened 2022-06-05
make first person challenge set players fov to 90

jhh8:


BenLubar:
We could also change the value in newmapsettings.

jhh8:
you mean fov_desired? how would that work?

BenLubar:
`fov_desired` is clamped to between 20 and 75 in third person and 20 and 120 in first person, so setting `fov_desired` in newmapsettings.cfg to 90 would only change the fov in first person

jhh8:
If we change newmapsettings now it will override it for everyone after the update right? Then i guess we can change fov to 90 internally in transition to asw controls 0

Dumbstep:
Maybe it can be added to the advanced settings

BenLubar:
We could add it to advanced settings (and mark it as FCVAR_ARCHIVE and remove it from newmapsettings), but I'd like to have sliders implemented for advanced settings if we do that.

BenLubar:
> If we change newmapsettings now it will override it for everyone after the update right? Then i guess we can change fov to 90 internally in transition to asw controls 0

the clamped version of fov_desired is recomputed when asw_controls changes, so anything between 75 and 120 will only affect first person

jhh8:
i mean the newmapsettings.cfg will override for everyone and people will lose their configs

BenLubar:
I'd like to get rid of the contents of newmapsettings.cfg eventually - it's weird to have default settings shipped as a config file.

BenLubar:
We could do something with FileProperties in the depot: https://partner.steamgames.com/doc/sdk/uploading#5

I'm not sure if there's any way to prevent the first update that adds that flag from resetting the file, though.

Nyubs:
lul that fake anticheat reacted and admin MANUALLY perma banned me, cause i had in newmapsettings fov desired 120 to watch a cheater.
and it was after the update , which limited it from the default client. so i hope that bad programmer of your anticheat fixed banning for using commands that are not even working anymore.
that would be exacty how you get banned from that admin, using not working FOV in third person. so need make new command or talk to that ADMIN :3

PR #78, opened 2022-06-12
Make asw_player_avoidance a server-side replicate cvar

Dumbstep:
Let the server control marine avoidance


BenLubar:
why not rd_marine_avoidance_override, defaulting to -1?

Dumbstep:
> how about rd_marine_avoidance_override, defaulting to -1?

Okay, I'll change it

element109:
Why not just move cvar asw_player_avoidance to server and replicate it? Why do you need 2 cvar for the same thing?

PR #83, opened 2022-06-13
Controller Improvements

element109:


element109:
I set the default for rd_controller_analog_radius_min to 0.5 because 0.3 will automatically blow you up with gl.

PR #108, opened 2022-06-18
Added Workshop Frame localization

Blueberryy:


BenLubar (changes requested):
Can you remove the change to basemodui_english.txt here and make it on the translation repo instead? The one in this repo is in UTF-16LE, so changing it like this will cause the game to fail to load it.

Blueberryy:
Just did. here

PR #127, opened 2022-06-22
fix not translated code

Blueberryy:


src/game/client/swarm/vgui/nb_weapon_detail.cpp
```diff
@@ -182,7 +182,7 @@ void CNB_Weapon_Detail::UpdateLabels( CASW_WeaponInfo *pWeaponData )
 
 			if ( flDamage <= 0.0f )
 			{
-				m_pValueLabel->SetText( L"#asw_weapon_altfire_NA" );
+				g_pVGuiLocalize->Find( "#asw_weapon_altfire_NA" );
```
BenLubar, line 185: go back to `SetText` but remove the `L`

Blueberryy: Just did

PR #139, opened 2022-06-29
Maintain assumption of possibly no vm

element109:


BenLubar:
Can asw_challenge_thinker even be created when there's no vscript VM?

PR #146, opened 2022-07-03
Filter dedicated servers from lobby list.

element109:
This needs to be tested. Issue #147


BenLubar:
Pretty sure `UTIL_RD_UpdateCurrentLobbyData` only works for the lobby owner, which is never a dedicated server. I think there's already a key in the lobby data for whether it's a listen or dedi server.

BenLubar:
> I tried this from the client it does not work either: `SteamMatchmaking()->GetLobbyGameServer( lobby, &serverIP, &serverPort, &serverID )`

yeah, Source Engine's matchmaking.dll uses a weird internal protocol to distribute game server connection info

probably designed before the Steam matchmaking API was a thing

BenLubar:
`options/server` should be correctly set to `dedicated` for lobbies created by beta clients when the next beta update goes out.

PR #180, opened 2022-08-02
Added more translateable stuff and missing objective reported by Deerlord

Blueberryy:
Thanks to @Deerlord0 for reporting missing objective description.


BenLubar (changes requested):
Ok, a few things:

- Not sure about translating the example map as the intention is for it to be viewed in Hammer, so the translation won't actually get processed for people who are looking at it.
- worldspawn strings aren't translated by the game, and they aren't *used* by anything, so I'd rather just blank them.
- The Accident 32 and Forgotten Factory changes are good 👍

Blueberryy:
> Ok, a few things:
> 
> * Not sure about translating the example map as the intention is for it to be viewed in Hammer, so the translation won't actually get processed for people who are looking at it.
> * worldspawn strings aren't translated by the game, and they aren't _used_ by anything, so I'd rather just blank them.
> * The Accident 32 and Forgotten Factory changes are good 👍

just reverted my things for worldspawn

PR #214, opened 2022-08-23
add queen sounds from city17

jhh8:
queen sounds from city 17 campaign


geotavros:
Make sure the content is not copyright protected. Otherwise RD risks being taken down from Steam.

PR #215, opened 2022-08-23
Lobby player info to match dedicated server.

element109:
This will avoid privacy issue as well.

### DIFF
--- a/src/game/client/c_playerresource.cpp
+++ b/src/game/client/c_playerresource.cpp
@@ -114,7 +114,13 @@ void C_PlayerResource::OnDataChanged(DataUpdateType_t updateType)
 			player_info_t info;
 			if ( IsConnected( slot ) && engine->GetPlayerInfo( slot, &info ) && !info.fakeplayer && !info.ishltv && !info.isreplay )
 			{
-				playerIDs.AppendFormat( "%s%08x", playerIDs.Length() ? "," : "", info.friendsID );
+				int index = engine->GetPlayerForUserID( info.userID );
+				int score = g_PR->GetPlayerScore( index );
+				CBasePlayer* pPlayer = UTIL_PlayerByIndex( index );
+				if ( pPlayer )
+				{
+					playerIDs.AppendFormat( "\"%s\"|%d|%f", playerIDs.Length() ? "," : "", pPlayer->GetPlayerName(), score, pPlayer->GetTimeBase() );
+				}
 			}
 		}
 
--- a/src/game/client/c_playerresource.cpp
+++ b/src/game/client/c_playerresource.cpp
@@ -127,7 +127,7 @@ void C_PlayerResource::OnDataChanged(DataUpdateType_t updateType)
 					{
 						playerIDs.AppendFormat( "%02x", *psz );
 					}
-					playerIDs.AppendFormat( "|%d|%f", score, pPlayer->GetTimeBase() );
+					playerIDs.AppendFormat( "|%d|%f", score, gpGlobals->curtime - pPlayer->GetConnectionTime() );
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


BenLubar:
Display names can contain quotes, so we need to escape them somehow.

src/game/client/c_playerresource.cpp
```diff
@@ -114,7 +114,13 @@ void C_PlayerResource::OnDataChanged(DataUpdateType_t updateType)
 			player_info_t info;
 			if ( IsConnected( slot ) && engine->GetPlayerInfo( slot, &info ) && !info.fakeplayer && !info.ishltv && !info.isreplay )
 			{
-				playerIDs.AppendFormat( "%s%08x", playerIDs.Length() ? "," : "", info.friendsID );
+				int index = engine->GetPlayerForUserID( info.userID );
+				int score = g_PR->GetPlayerScore( index );
+				CBasePlayer* pPlayer = UTIL_PlayerByIndex( index );
+				if ( pPlayer )
+				{
+					playerIDs.AppendFormat( "%s\"%s\"|%d|%f", playerIDs.Length() ? "," : "", pPlayer->GetPlayerName(), score, pPlayer->GetTimeBase() );
```
BenLubar, line 122:
```suggestion
					if ( playerIDs.Length() )
					{
						playerIDs.Append( ',' );
					}
					playerIDs.Append( '"' );
					for ( const char *psz = pPlayer->GetPlayerName(); *psz; psz++ )
					{
						if ( *psz == '\\' || *psz == '"' )
						{
							playerIDs.Append( '\\' );
						}
						playerIDs.Append( *psz );
					}
					playerIDs.AppendFormat( "\"|%d|%f", score, pPlayer->GetTimeBase() );
```

mithrand0:
Can you use the same logic as KeyValues uses for escaping?

https://github.com/ValveSoftware/source-sdk-2013/blob/master/sp/src/tier1/KeyValues.cpp#L707

BenLubar:
The quotation marks and escaping of `\` and `"` handle marking the boundaries of the name.

mithrand0 (approved):

src/game/client/c_playerresource.cpp
```diff
@@ -114,7 +114,22 @@ void C_PlayerResource::OnDataChanged(DataUpdateType_t updateType)
...
+				if ( pPlayer )
+				{
+					if ( playerIDs.Length() )
+					{
+						playerIDs.Append( ',' );
+					}					
+					for ( const char *psz = pPlayer->GetPlayerName(); *psz; psz++ )
+					{
+						playerIDs.AppendFormat( "%02x", *psz );
+						playerIDs.Append('-');
```
BenLubar, line 129: Since the bytes are each 2 hex characters, we don't need separators between the bytes.

```suggestion
```

mithrand0:
The site is prepared for this patch: https://github.com/mithrand0/reactive-drop-tools-site/commit/d6e6cb1866a7d8aa350898057a0181735fe39fef

src/game/client/c_playerresource.cpp
```diff
@@ -114,7 +114,21 @@ void C_PlayerResource::OnDataChanged(DataUpdateType_t updateType)
...
+				{
+					if ( playerIDs.Length() )
+					{
+						playerIDs.Append( ',' );
+					}					
+					for ( const char *psz = pPlayer->GetPlayerName(); *psz; psz++ )
+					{
+						playerIDs.AppendFormat( "%02x", *psz );
+					}
+					playerIDs.AppendFormat( "|%d|%f", score, pPlayer->GetTimeBase() );
```
BenLubar, line 130: one last thing: GetTimeBase returns the number of seconds between the map spawning and the player last being simulated, which isn't what we want here

so that we don't have to update the lobby every second, we should set this to an absolute timestamp of when the player joined the server

problem is, I'm not sure how to get that on the client

BenLubar: worst case scenario we can add a netprop to player

BenLubar: `gpGlobals->frametime` is the amount of time it took the previous frame to run.

Since the lobby owner is generally the first person to connect to a game (in the case of a listen server, the host), setting `m_nConnectionTime = std::time( NULL );` in the C_BasePlayer constructor could work. Might need to `#include <ctime>`

PR #218, opened 2022-08-23
Add ScriptTempEnts Class and Functions

Rayman1103:
- Add new VScript class **TempEnts** which allows creating Temporary Entities on clients.
- Add new global VScript functions to use when creating Temporary Entities.

**TempEnts**

_Queue a temp entity for transmission on a client from a passed table of SendProp data.
Passing null for a player will playback Temp Entity on all clients._
`void Create( CBasePlayer player, string name, float delay, table )`

_Fills in a passed table with all SendProps and their types for the temp entity_
`void GetPropTypes( string name, table )`

_Fills in a passed table with the names of all temp entities_
`void GetNames( table )`

**Global**

_Precaches a particle material_
`int PrecacheParticleSystem( string name )`

_Converts a previously precached material into an index_
`int GetParticleSystemIndex( string name )`

_Converts a previously precached material index into a string_
`string GetParticleSystemNameFromIndex( int index )`

_Precaches an effect_
`void PrecacheEffect( string name )`

_Converts a previously precached effect into an index_
`int GetEffectIndex( const char *name )`

_Converts a previously precached effect index into a string_
`string GetEffectNameFromIndex( int index )`

_Get decal index from a string_
`int GetDecalIndexForName( string name )`

### DIFF
--- a/src/game/server/vscript_server.cpp
+++ b/src/game/server/vscript_server.cpp
@@ -22,6 +22,7 @@
 #include "ai_link.h"
 #include "ai_basenpc.h"
 #include "inetchannelinfo.h"
+#include "decals.h"
 #ifdef _WIN32
 #include "vscript_server_nut.h"
 #endif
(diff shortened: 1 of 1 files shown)


jhh8:
epic

PR #287, opened 2022-09-12
vcsripts: add new functions for player associated with tonemap

jhh8:
SetAutoExposure( float min, float max ) - Sets custom autoexposure (simply - brightness) for the player

SetBloomScale( float scale ) - Sets a custom bloomscale for the player

SetTonemapRate( float rate ) - Sets a custom tonemaprate (rate for autoexposure adjustment) for the player

TODO: Add a function to reset changes in tonemap back to map's default settings

### DIFF
--- a/src/game/client/c_env_tonemap_controller.cpp
+++ b/src/game/client/c_env_tonemap_controller.cpp
@@ -5,6 +5,7 @@
 //=============================================================================
 
 #include "cbase.h"
+#include "c_user_message_register.h"
 
 // NOTE: This has to be the last file included!
 #include "tier0/memdbgon.h"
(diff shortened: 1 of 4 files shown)


BenLubar:
@jhh8 does https://github.com/ReactiveDrop/reactivedrop_public_src/commit/299726793267df4a02180da2bf1a70c611204d31 do what you wanted to accomplish with this?

jhh8:
Makes it easier for mappers but not for people that make challenges. Though if it works correctly then people making challenges can reset the tonemap effects done with my vscripts by doing set tonemap controller to the tonemap controller found on the map with your vscript

BenLubar:
Do we need an easier way to create an env_tonemap_controller that is a copy of another one's settings? Or an easier way to find the master tonemap controller from VScript? VScript can definitely already create a new env_tonemap_controller and give it a name and send it inputs.

jhh8:
i want to make a feature in a challenge where lower health means darker vision, simple setautoexposure will be much more convenient than creating a lot of tonemap controller handles and choosing between them

jhh8:
though after testing a bit it seems like cant go into negatives so on some maps it would be inpossible to make darker than it already is, only solution would be configurable fog then. but these functions still might be convenient to use, maybe someone crazy will make it more brighter with less health someone has

BenLubar:
Let me know if you still need more tonemap vscript functions.

PR #297, opened 2022-09-19
HARVESTER: closer to previous behaviour + a convar

jhh8:
after the previous commit realised that it would be very inconsistent a.k.a hard to predict because it could lead to situations where xenomytes wouldn't be released when the nade hit was not direct

optimal but not a clean way to check if it was a direct nade hit ( info.GetDamage() > 135.0f ) ensures that the explosion which killed the harvesters was a direct nade hit, moree consistent behaviour

+a convar to make xenomytes always be released if some challenges wanna use it to make it a bit harder

### DIFF
--- a/src/game/server/swarm/asw_harvester.cpp
+++ b/src/game/server/swarm/asw_harvester.cpp
@@ -42,6 +42,7 @@ ConVar asw_harvester_new( "asw_harvester_new", "1", FCVAR_CHEAT, "If set to 1, u
 ConVar asw_harvester_spawn_height( "asw_harvester_spawn_height", "16", FCVAR_CHEAT, "Height above harvester origin to spawn xenomites at" );
 ConVar asw_harvester_spawn_interval( "asw_harvester_spawn_interval", "1.0", FCVAR_CHEAT, "Time between spawning a harvesite and starting to spawn another" );
 ConVar rd_harvester_health( "rd_harvester_health", "200", FCVAR_CHEAT, "Health of the harvester" );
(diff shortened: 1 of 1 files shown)


src/game/server/swarm/asw_harvester.cpp
```diff
@@ -42,6 +42,7 @@ ConVar asw_harvester_new( "asw_harvester_new", "1", FCVAR_CHEAT, "If set to 1, u
 ConVar asw_harvester_spawn_height( "asw_harvester_spawn_height", "16", FCVAR_CHEAT, "Height above harvester origin to spawn xenomites at" );
 ConVar asw_harvester_spawn_interval( "asw_harvester_spawn_interval", "1.0", FCVAR_CHEAT, "Time between spawning a harvesite and starting to spawn another" );
 ConVar rd_harvester_health( "rd_harvester_health", "200", FCVAR_CHEAT, "Health of the harvester" );
+ConVar rd_harvester_always_release_xenomites( "rd_harvester_always_release_xenomites", "0", FCVAR_NONE, "If set, harvesters will always release xenomites after death" );
```
BenLubar, line 45:
```suggestion
ConVar rd_harvester_always_release_xenomites( "rd_harvester_always_release_xenomites", "0", FCVAR_CHEAT, "If set, harvesters will always release xenomites after death" );
```

Challenge convars should be FCVAR_CHEAT.

BenLubar (changes requested):

PR #299, opened 2022-09-20
New RGB Blending for Chat Text

ag3killstreak:
Originally, I was just messing around with this code not planning on uploading it, but peeps seemed to like it so I Improved it and added more functionality.

This new addition adds blending for chat text. It works similar to the previous RGB system and uses two of the same 0x08 characters to send data to the client. This is followed by an encoded blend type to tell the client which blend mode we are using. These new blending functions would allow a new colorful blend of colors for modders and server owners to use.

The first is the basic blend that blends two sets of RGB colors. The blend starts at the point after the encoded sequence finishes and ends after (either A. The end of the text B. Another color char is detected.)

Example:
ClientPrint(null, 3, TextColorBlend(255, 155, 0, 255, 255, 0) + "This is a test to show blended text using new color blend function. As you can see the text blends from color1 to color2 as it gets closer to the tail end of the text.");

The next is an inversion blend that blends a single RGB input to it's inverted color. Works the same way as normal blend only it doesn't need to send extra 3 bytes for second RGB input.

Example:

ClientPrint(null, 3, TextColorBlendInvert(255, 255, 125) + "This is a test to show inverted blend text. It blends into the inverted color for the supplied RGB input.");

The next is a blend of 3 colors. Works the same only has an additional RGB color.

Example:

ClientPrint(null, 3, TextColorBlend3(255, 55, 55, 255, 255, 255, 55, 55, 255) + "This is a test to show the 3 color blend text. It blends into the 3 colors for the supplied RGB inputs.");

The next is a color blend cycle. The color blend cycle's first argument is the length of a transition followed by two RGB inputs. The blend cycle allows repeating a color pattern. The pattern will jump to color 1 after it reaches the end of it's provided length.

Example:
ClientPrint(null, 3, TextColorBlendCycle(5, 122, 122, 255, 155, 155, 155) + "This is a test to show the color cycle blend text. It cycles the blend through color 1 to color 2 using the inputted length to control how long the blend is.");

The last is a color blend smooth cycle. It works the same as normal blend cycle only it smoothly returns to color 1.

Example:
ClientPrint(null, 3, TextColorBlendSmoothCycle(25, 255, 155, 155, 155, 155, 255) + "This is a test to show the color smooth cycle blend text. It cycles a smooth blend through color 1 to color 2 and back to color 1 using the inputted length to control how long the blend is.");

### DIFF
--- a/src/game/client/hud_basechat.cpp
+++ b/src/game/client/hud_basechat.cpp
@@ -52,7 +52,7 @@ Color g_ColorGreen( 153, 255, 153, 255 );
 Color g_ColorDarkGreen( 64, 255, 64, 255 );
 Color g_ColorYellow( 255, 178.5, 0.0, 255 );
 Color g_ColorGrey( 204, 204, 204, 255 );
-Color g_ColoPurple( 160, 115, 205, 255 );
+Color g_ColorPurple( 160, 115, 205, 255 );
 
 static const char *gBugPriorityTable[] = {
 	"TODAY", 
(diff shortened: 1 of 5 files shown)


jhh8:
constexpr int CLAMP_COLOR( int value )
{
	return value < 0 ? 0 : value > 255 ? 255 : value;
}

R = CLAMP_COLOR(R);

BenLubar:
Does this work correctly with surrogate pairs?

ag3killstreak:
> surrogate pairs

Testing multibyte in vscripts was annoying as .nut doesn't support anything other than ANSI.

You can test it by inserting the following in console:
script ClientPrint(null, 3, TextColorBlend3(255, 15, 15, 255, 255, 255, 100, 100, 255) + "こんにちは、このテキストは日本語ですこんにちは、このテキストは日本語ですAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA");

Result is:
https://cdn.discordapp.com/attachments/774748397901971476/1022271558380703784/blend_multibyte_characters_test1.jpg

I programmed it to index based on wchar_t object so I believe it should work properly on all multibyte characters.

ag3killstreak:
> > surrogate pairs
> 
> Testing multibyte in vscripts was annoying as .nut doesn't support anything other than ANSI.
> 
> You can test it by inserting the following in console: script ClientPrint(null, 3, TextColorBlend3(255, 15, 15, 255, 255, 255, 100, 100, 255) + "こんにちは、このテキストは日本語ですこんにちは、このテキストは日本語ですAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA");
> 
> Result is: https://cdn.discordapp.com/attachments/774748397901971476/1022271558380703784/blend_multibyte_characters_test1.jpg
> 
> I programmed it to index based on wchar_t object so I believe it should work properly on all multibyte characters.

Actually further testing shows wchar_t is returning segments of 2 bytes at a time on a 4 byte char so code needs a rework.

PR #302, opened 2022-09-20
New translateables related to keybinding

Blueberryy:


src/game/client/cdll_util.cpp
```diff
@@ -830,7 +830,7 @@ void UTIL_ReplaceKeyBindings( const wchar_t *inbuf, int inbufsizebytes, wchar_t
 				const char *key = engine->Key_LookupBindingEx( *binding == '+' ? binding + 1 : binding, nSlot );
 				if ( !key )
 				{
-					key = IsX360() ? "" : "< not bound >";
+					key = IsX360() ? "" : "#GameUI_KeyNames_Not_Nound_lowercase";
```
BenLubar, line 833: typo

BenLubar:
I think it would be fine if there was only one translation key for both of them, for either `< not bound >` or `<NOT BOUND>` depending on which looks better

Blueberryy:
> I think it would be fine if there was only one translation key for both of them, for either `< not bound >` or `<NOT BOUND>` depending on which looks better

You can change it if you want meow

PR #313, opened 2022-09-25
update the readme

mithrand0:
I updated the readme to reflect how things are already. The change is not a change request, it just describes the situation better to what it already is.

I leave the merge to someone else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,8 +48,7 @@ This way we will be able to see which changes did you introduce to the content f
 
 Changes can be submitted as pull requests.
 
-A pull request should be as granular as possible and only change one thing at a time. If changes are related or more complex, it would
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,8 +5,9 @@
 ### What is this repository for? ###
 Having access to the source code you can:
 
-* Add new weapons, NPCs, fix bugs and generally add new features to the game and submit them as pull requests 
-* Create mods. Both server side and client side mods. Or fully standalone mods
```
geotavros, line 9: This shouldn't be removed. We still allow creating mods from our source. If you want to make it clear that server and client mods shouldn't be used to cheat leaderboards or cheat on ranking servers then this can be added as a separate line.

BenLubar:
Maybe note the convar that is required to use a client mod (`rd_check_clientdll_consistency`)

PR #318, opened 2022-09-27
OnSetTransmit VScripts Callback hook

ag3killstreak:
I was requested to make this.

Adds vscript callback OnSetTransmit(client, entity)

This callback allows blocking entity transmits to specific clients.

Example:
function OnSetTransmit(client, entity)
{
	local classname = entity.GetClassname();

	if (classname == "asw_drone")
	{
		return 0;
	}
	return 1;
}

### DIFF
--- a/src/game/server/baseentity.cpp
+++ b/src/game/server/baseentity.cpp
@@ -4200,6 +4200,57 @@ void CBaseEntity::SetTransmit( CCheckTransmitInfo *pInfo, bool bAlways )
 	if ( pInfo->m_pTransmitEdict->Get( index ) )
 		return;
 
+	//AG3_KILLSTREAK added code:
+	int iVScriptsShouldTransmit = -1;
+	if (index>0 && !m_bIgnoreVScriptTransmit && g_pScriptVM)
+	{
+		edict_t* clientEdict = pInfo->m_pClientEnt;
+		int clientIndex = ENTINDEX(clientEdict);
+		CBaseEntity* clientEntity = UTIL_EntityByIndex(clientIndex);
+
+		if (HSCRIPT hFunction = g_pScriptVM->LookupFunction("OnSetTransmit"))
+		{
+			ScriptVariant_t newResult;
+			ScriptStatus_t nStatus = g_pScriptVM->Call(hFunction, NULL, true, &newResult, ToHScript(clientEntity), ToHScript(this));
+			if (nStatus != SCRIPT_DONE)
+			{
+				DevWarning("OnSetTransmit VScript function did not finish!\n");
+			}
+			else
+			{
+				newResult.AssignTo(&iVScriptsShouldTransmit);
+			}
+			g_pScriptVM->ReleaseFunction(hFunction);
+		}
+		else if (g_pScriptVM->ValueExists("g_ModeScript"))
+		{
+			ScriptVariant_t hModeScript;
+			if (g_pScriptVM->GetValue("g_ModeScript", &hModeScript))
+			{
+				if (HSCRIPT hFunction = g_pScriptVM->LookupFunction("OnSetTransmit", hModeScript))
+				{
+					ScriptVariant_t newResult;
+					ScriptStatus_t nStatus = g_pScriptVM->Call(hFunction, hModeScript, true, &newResult, ToHScript(clientEntity), ToHScript(this));
+					if (nStatus != SCRIPT_DONE)
+					{
+						DevWarning("OnSetTransmit VScript function did not finish!\n");
+					}
+					else
+					{
+						newResult.AssignTo(&iVScriptsShouldTransmit);
+					}
+					g_pScriptVM->ReleaseFunction(hFunction);
+				}
+				g_pScriptVM->ReleaseValue(hModeScript);
+			}
+		}
+	}
+	if (iVScriptsShouldTransmit == 0)
+	{
+		return;
+	}
+	//END OF ----------------
+
 	CServerNetworkProperty *pNetworkParent = NetworkProp()->GetNetworkParent();
 
 	pInfo->m_pTransmitEdict->Set( index );
--- a/src/game/server/baseentity.h
+++ b/src/game/server/baseentity.h
@@ -749,6 +749,8 @@ public:
 	string_t m_iGlobalname; // identifier for carrying entity across level transitions
 	string_t m_iParent;	// the name of the entities parent; linked into m_pParent during Activate()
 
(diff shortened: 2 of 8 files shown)


jhh8:
what would this be used for? this is very dangerous

ag3killstreak:
> what would this be used for? this is very dangerous

Dangerous in what way?

Sourcemod has had this functionality for a long time.

public Action:Hook_SetTransmit(entity, client) ;

jhh8:
not transmitting entities like asw_game_resource crashes clients, not transmitting some other entities (couldnt figure out which) crashes clients after mission restart. this function gets called 2000 times a second per client, also having an error like for example local a = "a" / 2; freezes the server forever. there might be more reasons that someone familiar with client prediction or other stuff might find

jhh8:
the example you posted in the description can be achieved by setting drones modelscale to 0 making them invisible

ag3killstreak:
> not transmitting entities like asw_game_resource crashes clients, not transmitting some other entities (couldnt figure out which) crashes clients after mission restart. this function gets called 2000 times a second per client, also having an error like for example local a = "a" / 2; freezes the server forever. there might be more reasons that someone familiar with client prediction or other stuff might find

I actually forgot to include the asw_debrief_stats.cpp file in the pull request. Thanks for pointing that out.  "asw_game_resource" does indeed crash which is why I added the boolean for skipping it in vscripts check.

The function does get fired quickly, yes.  It functions the same way it does in sourcemod's callback.

The purpose of the script is actually to desync clients entities from one another to allow for unique changes.
For instance, The survivor legs plugin for l4d2 uses this callback to show parented leg model for specific client and not others.

I tested local a = "a" / 2; and it does not even dent the server frame rate so unless if you're talking about srcds, Idk what you're talking about.

I couldn't find any other entities that crashed other than that one which was caused by the client waiting for network input before creating the entity and populating the pointer.

The only other thing I noticed is an assert for the bucket controller, but that does not cause any crashes that I'm aware of.

My vs2022 corrupted its github link so I have to manually copy the files into the directory I use to upload. I'll do a search for "m_bIgnoreVScriptTransmit = true;" to see if I missed any files when I get back.

I tested it before doing the pull request with

function OnSetTransmit( client, entity) 
{
return 0;
}

It desynced, it gave some asserts, but it didn't crash, which is by design.

ag3killstreak:
> the example you posted in the description can be achieved by setting drones modelscale to 0 making them invisible

I'd just use renderfx if I was going to make them invisible. I included that code as an example to test.

jhh8:
function OnSetTransmit( client, entity )
{
	if ( entity.GetClassname() == "asw_game_resource" || entity.GetClassname() == "asw_campaign_save" )
		return 1;
	
	return 0;
}

this is crashing the client after a mission restart with my tests on landing bay map

ag3killstreak:
Ok, I found 3 more files I didn't copy over.

BenLubar:
What problem are you trying to solve with this? This seems like a low-level networking solution to a high-level mission design problem, which I don't think is the right way to go.

ag3killstreak:
> What problem are you trying to solve with this? This seems like a low-level networking solution to a high-level mission design problem, which I don't think is the right way to go.

The OnSetTransmit hook is designed to allow recieved client data to be unique from the other clients. For instance, A visible entity that is invisible on the others. First person view models that should only be shown in 3rd person, or vis versa. Model Index changing for things like unique scoreboards.

It's a more advanced tool for modders to do some things that can be done on sourcemod, but can't be done using vscripts.

jhh8:
what about a way to remove/add an entity into m_pTransmitEdict which would stop its transmission, should be much less expensive. it looks like it can be done for any client seperately

ag3killstreak:
> what about a way to remove/add an entity into m_pTransmitEdict which would stop its transmission, should be much less expensive. it looks like it can be done for any client seperately

Would have to see.

I believe some plugins alter netprops before they are sent so this solution would be nonfunctional for them.

jhh8:
the idea is making the entity transmition blocked with an earlier return from the function without the need of a hook

image

so it should work exactly the same as your current method

ag3killstreak:
> the idea is making the entity transmition blocked with an earlier return from the function without the need of a hook
> 
> image
> 
> so the result should be exactly the same as with your method except a lot cheaper

Maybe it could be done that way if the callback is moved to before any SetTransmits are called for a client (depending on the code structure).  The hook isn't just to block the network input or your solution would probably work fine.

jhh8:
Im not understanding you here, the second solution would be either as fast or faster. With your solution the hook starts working after the vscripts are initialized and so the OnSetTransmit hook is parsed when the nut file is run. 

In the second solution, in the nut file you can put the code outside of a function and it would be parsed immediately when vscripts are initialized and the file is run

 so same speeds or might be even faster since the code is not in a function and it might take a milisecond longer for the game to initialize your function

mithrand0:
> What problem are you trying to solve with this? This seems like a low-level networking solution to a high-level mission design problem, which I don't think is the right way to go.

I agree with Ben. SetTransmit has the potential to cause crashes, lag, and other symptoms if incorrectly used.

> It's a more advanced tool for modders to do some things that can be done on sourcemod, but can't be done using vscripts.

I searched, but all I could find are some threads on Allied Mods with people having issues with it. I probably searched wrong. Do you have an example of such plugin(s) for us?

MartLegion:
Some use cases to SetTransmit in Sourcemod (L4D based and client/team-specific settings):
- Control if an entity can be seen while in first/third-person view; (usually the ones attached to the head)
- Hide items in versus matches (like powerups/weapons) so the opposite team doesn't know where and which items were spawned;

mithrand0:
I tried search for it, but couldn't find the mods you mention. Do you have any links to source code of the mentioned mods?

Regarding the use cases, implementing team/player based visibility sounds more correct here, instead of changing transmit properties.

Shadowysn:
Maybe this could be made to only exclusively work for cosmetic entities (props/brushes) since it seems to be the use case for the majority of sourcemod plugins using SetTransmit.

Silver's L4D2 hats attached to players use SetTransmit to be hidden only to the owner to avoid blocking first-person view, so it's not something just team-based visibility can solve: https://forums.alliedmods.net/showthread.php?p=1441080

I wanted to reply here since I did bring this up as a nice feature to have for L4D2's VScript and it prompted this short-lived pull request, and it also showed the further-reaching consequences I didn't really think about, but in a way I was too intimidated to really respond for years as the people here actually work in-depth on source code and I only really did surface-level sourcemod plugins.

Issue #163, opened 2022-07-18
Disk activity causing lag


mithrand0:
Some disk activity is causing extreme lag on my dedicated servers.

The issue occurs when Windows does regular maintenance. Disk activity spikes, causing lags for players at that moment. When the activity is over, the game runs like normal. Cpu is not very high during that time, more than enough ram is free.

This surprises me a bit, because the `logs`, `save` and `console.log` are put on a ramdisk. The running server could lag because slower disk reads during a map load, but this should not impact a running game.

Maybe I missed a file that is actively used or needed during a running map?

BenLubar:
I can't think of any reason the server would be reading/writing files on disk other than logs during gameplay. The files in `save` get read/written only while the game is in briefing/debriefing state. The only file writing I can think of outside of `save` and `logs` would be `ISteamGameServerStats` calls, but those all happen at very predictable times.

FurtadoPires:
> Disk activity spikes, causing lags for players at that moment

You should check for those reasons too:

1 - Windows may be using swap due to lack of ram, which could make things slow specially if you're using HD instead of a SSD.

2 - Your disk might have badblocks, and the slowness may be windows trying to read/write on those bad sectors. I personally don't know any reliable windows software to check for disk badblocks, but I know an excellent method to verify using a pen drive with linux. I could teach you how in case you might want to check.

mithrand0:
> 1 - Windows may be using swap due to lack of ram, 
Like written above, server has enough ram free (30% used), system shows no use of paging

> 2 - Your disk might have badblocks <...>
Can be ruled out, it's hardware raid that does scheduled disk patrols (also, the timing of that does not match the lags)

The issue is not slow disk usage or i/o. The issue is whenever the disk is used, there is a (micro)stutter in the game. The more disk i/o, the more it becomes noticeable. 

> Ben: I can't think of any reason the server would be reading/writing files on disk other than logs during gameplay. 

Exactly this. Trying to trace it the process locally to see if and when it does access the file system.

PR #326, opened 2022-10-03
[needs-testing] do not save the game (to disk) on marine death or revive.

mithrand0:
related to #163


BenLubar:
Is this happening so often that it's causing noticeable lag?

PR #327, opened 2022-10-03
res4: fix lags associated with phys_convert

### LINKED ISSUE #133: Lag on mission with queen
Servers are dropping in sv whenever the official mission with queen is played. The lag starts when the queen touches a specific container. If you close the door after, it'll lag as well.

Images of the specific spot:
https://steamuserimages-a.akamaihd.net/ugc/1877453375037615622/BB3B83359483BC8CF86E996A6585797D0B9996EA/
https://steamuserimages-a.akamaihd.net/ugc/1877453672148515625/510C955935455E58D608A9AC453626F248364DCB/

The issue is hard to reproduce and weirdly not always visible on self-hosted servers, but on ASRD Germany it's fully reproducible. 

It might have something to do with collision issues, but unsure why some servers have issues, while others have (randomly) not.

jhh8:
phys_convert very buggy... better use prop physics override with disabled motion flag


jhh8:
fixes #133

mithrand0:
We tested the fix, it fixes the issue mentioned in #133

PR #328, opened 2022-10-04
[do-not-merge] update the default values 

mithrand0:
Update the default values from the 2010 dial up era to something that is more modern. 

New default values for calculations are taken from CS:GO and are best practices used by CS:GO ESL servers. If something mismatches, feedback is appreciated, as long as it specifies the calculation and reasons behind it.

Futhermore, I would like to test these with a broader group. If you are interested in joining the test group, feel free to join https://s.team/chat/XcvzBOy7 (channel: Lobby testing).

### DIFF
--- a/src/game/client/swarm_sdk_client.vcxproj
+++ b/src/game/client/swarm_sdk_client.vcxproj
@@ -512,6 +512,7 @@ if exist ..\..\devtools\bin\postbuild.cmd ..\..\devtools\bin\postbuild.cmd clien
     <ClCompile Include="..\shared\predictioncopy_test.cpp" />
     <ClCompile Include="..\shared\props_shared.cpp" />
     <ClCompile Include="..\shared\ragdoll_shared.cpp" />
+    <ClCompile Include="..\shared\rd_defaults.cpp" />
     <ClCompile Include="..\shared\rope_helpers.cpp" />
     <ClCompile Include="..\shared\saverestore.cpp" />
     <ClCompile Include="..\shared\sceneentity_shared.cpp" />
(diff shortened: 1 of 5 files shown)


BenLubar:
you can mark the PR as a draft

also, should this be made part of the cvar hacks file that already exists?

mithrand0:
Converted it to draft.

Yes, we can merge this with rd_convar_hacks, I'll see to it.

geotavros:
I'd be very very careful with network rates changes. RD uses both p2p connections and dedicated servers. If we don't know what is the guaranteed stable upload speed of the player we shouldn't enforce ESL speeds on them. Some players have 1 MBit/s upload speed. I've heard people form Australia, New Zeeland and even US have like 1-5 mbit upload speed because of ISP monopolies or too expensive plans. I didn't do a research though. Some players play on 3G or 4G. A lot of players play through wi-fi, and their router is shared among all family members, where their upload speed may jump from 3 to 25 MBit/s.

Note that CSGO is not a p2p game, at least from what I know. You don't host a server from your game. Dedicated servers are hosted on machines with good cable internet, or else nobody will play on laggy wi-fi servers.

When doing any networking changes it needs to be very well tested. With various speed limits scenarios in mind on the most network-heavy official maps. 

Only 1 of 100 new players will go and write a review or forum post that game is laggy, stutters and teleports. 99 will just quit and un-install.

mithrand0:
Agree this needs to be very careful implemented, I don't expect this PR to be merged, unless extensively tested before.

Regarding hosting, if a lobby host only has 1 Mbit upload, how would that work out with the current settings. What if they run 8 player brutal ASBI with stock settings. The choke would create lag already at current default settings.

geotavros:
1 Mbit player can only stably host a 4 player server. I have seen such scenarios when 6th player joins the game and everyone starts to have ~300 ping when they had ~50.

We can look how other games solve this. My assumption is that games try to use as least bandwidth as possible. For example, in l4d2 it's not possible to go higher than 30 updaterate, tickrate and server fps is locked to 30.

BenLubar:
I don't think we can have this enabled/disabled using a convar due to when it runs. I'd say it should be a command line argument to disable it, like how `-skiploadingworkshopaddons` is set up.

PR #338, opened 2022-10-06
New RGB custom laser sights

ag3killstreak:


ag3killstreak:
Whoops, I wasn't trying to submit this here yet.

I was trying to add it to the the other repository.

BenLubar:
You can mark this as a draft until it's ready to merge.

ag3killstreak:
> You can mark this as a draft until it's ready to merge.

No idea how to do that.

ag3killstreak:
nvm, Found it.

src/game/client/swarm/c_asw_concommands.cpp
```diff
@@ -1122,4 +1122,19 @@ CON_COMMAND( make_game_friends_only, "Changes access for the current game to fri
...
\ No newline at end of file
+}
+
+
+extern ConVar cl_asw_laser_sight_color;
+void ASW_ChangeLaser_f(const CCommand& args)
+{
+	if (args.ArgC() >= 4)
+	{
+		engine->ClientCmd(VarArgs("cl_changelaser %d %d %d", atoi(args[1]), atoi(args[2]), atoi(args[3])));
```
BenLubar, line 1133: this could be something that happens in a convar change callback if `engine->IsConnected()`

ag3killstreak: Is there a good way to know if client is currently in lobby screen?

Currently, I have code in "nb_main_panel.cpp" that detects changes, but only sends out updates every 250ms so it doesn't flood server everytime a new color is picked. 

I could do it that way, but I need a way to know if lobby screen is active so that it can handle the laser color convar changes without flooding.

ag3killstreak: `	if (m_LastNetworkSentColor != cl_asw_laser_sight_color.GetColor() && gpGlobals->curtime > m_lastColorNetworkUpdate + 0.25f)
	{
		SetChangingLaserColor(cl_asw_laser_sight_color.GetColor());
		m_LastNetworkSentColor = cl_asw_laser_sight_color.GetColor();
		m_lastColorNetworkUpdate = gpGlobals->curtime;
	}`

BenLubar: You could only update the convar when the mouse is released to reduce the amount of updates.

ag3killstreak: I added a method to the client player class that gets if briefing panel pointer is nonzero.

ag3killstreak:
I still haven't been able to get 8 players to test with full human team but all other tests are working.

ag3killstreak:
The resources directory will have to be changed before merging as I currently have it set up to include them as an addon.

BenLubar:
The release on the 27th will be bugfixes only, but this may be included in the next release after October.

src/game/client/swarm/c_asw_weapon.cpp
```diff
@@ -885,31 +914,103 @@ void C_ASW_Weapon::SimulateLaserPointer()
 	m_bLocalPlayerControlling = bLocalPlayer;
 }
 
-void C_ASW_Weapon::CreateLaserPointerEffect( bool bLocalPlayer, int iAttachment )
+int C_ASW_Weapon::GetFreeLaserIndex(bool bSetAsTaken)
 {
-	if ( m_pLaserPointerEffect )
+	int index = 0;
+	while (index < 8) //LS INDEX SIZE 8
```
BenLubar, line 920: There can be up to ASW_MAX_MARINE_RESOURCES marines on the map in, for example, deathmatch.

src/game/client/swarm/c_asw_weapon.h
```diff
@@ -14,6 +14,10 @@ class C_ASW_Marine;
 class CASW_WeaponInfo;
 class C_BreakableProp;
 
+static bool g_bLaserIndexInUse[8]{}; //LS INDEX SIZE 8
```
BenLubar, line 17: don't define a naked static variable in a header - this causes each file that includes the header to have a separate instance of the variable

you may want extern here and then the same variable without extern in a cpp file

src/game/server/swarm/asw_marine.cpp
```diff
@@ -634,6 +635,16 @@ CASW_Marine::CASW_Marine() : m_RecentMeleeHits( 16, 16 )
 
 	m_iPoisonHeal = 0;
 	m_flNextPoisonHeal = -1;
+
+
+	Vector vecCol = Vector(0, 0, 0);
+	Vector vecHSV = Vector(RandomFloat(0.0f, 360.0f), RandomFloat(0.0f, 255.0f), 255.0);
```
BenLubar, line 641: Instead of random, the default should be the color in g_rgbaStatsReportPlayerColors for the marine resource index

src/game/server/swarm/asw_marine.h
```diff
@@ -285,6 +285,8 @@ class CASW_Marine : public CASW_VPhysics_NPC, public IASWPlayerAnimStateHelpers,
 	virtual bool ShouldNotDistanceCull();	// BenLubar(deathmatch-improvements):
 	virtual Disposition_t IRelationType( CBaseEntity *pTarget );	// BenLubar(deathmatch-improvements):
 
+	CNetworkVar(Vector, m_vecCustLaserColor);
```
BenLubar, line 288:
```suggestion
	CNetworkVector(m_vecCustLaserColor);
```

also consider moving the variable to the marine resource

src/game/server/swarm/asw_player.h
```diff
@@ -244,6 +244,7 @@ class CASW_Player : public CBaseMultiplayerPlayer, public IASWPlayerAnimStateHel
 	// status of selecting marine/weapons in the briefing
 	CNetworkVar( int, m_nChangingMR );
 	CNetworkVar( int, m_nChangingSlot );
+	CNetworkVar(Vector, m_vecLobbyLaserColor );
```
BenLubar, line 247:
```suggestion
	CNetworkVector( m_vecLobbyLaserColor );
```

are laser colors per-player or per-marine?

src/game/shared/particle_property.cpp
```diff
@@ -200,6 +200,43 @@ CNewParticleEffect *CParticleProperty::Create( const char *pszParticleName, Part
 	return Create( pDef, iAttachType, iAttachmentPoint, vecOriginOffset, matOffset );
 }
 
+CNewParticleEffect* CParticleProperty::Colored_Create(const char* pszParticleName, ParticleAttachment_t iAttachType, int iAttachmentPoint, Vector colorMul, Vector vecOriginOffset, matrix3x4_t* matOffset)
```
BenLubar, line 203: not sure about this. particle effects are supposed to use control points to transmit information, not edit the particle effect definition at runtime.

BenLubar (commented):
initial review

src/game/server/swarm/asw_player.h
ag3killstreak: In lobby, I have it programmed for each human player to be able to set their own laser colors.

AI currently just choose a random color along the H and S spectrum.

ag3killstreak: In game, each marine has it's own color netprop.

src/game/shared/particle_property.cpp
ag3killstreak: The definition class data is inside the particle..lib file so I can't access it.

My work around was to create template lasers in the pcf file, one for each marine to use.

The definition uses a bunch of pointers so to create a new definition would require emulating the functions inside particle.lib.

A better solution would be to create a new particle definitions class as a clone, but all my attempts failed with conflicts in the particle.lib file.

The current process detects changes to the netprop on the client side, checks for a currently unused particle effect template index, and alters the existing particle definition template data.

Without source to particle.cpp, I'm kindof stuck unless you know of a better way to generate a definition clone.

The index is never networked, it is just a local variable to keep track of what is currently in use on the client's end.

BenLubar: Can we just add the color for the laser pointer as an input control point to the particle system and then set that? That's how the game handles, for example, icons on the floor for stuff like "weld this door"

ag3killstreak: Maybe?

Do you have an example of how and where that is done at so I can look into it?

ag3killstreak: You mean like this?

m_pLaserPointerEffect->SetControlPoint(20, Vector(255, 255, 255));

ag3killstreak: Doesn't work.

What indexes control "m_ColorMin" and "m_ColorMax"?

BenLubar: image

Open the game in tools mode via the SDK
weapon_laser_sight is in asw_weapon_fx.pcf (you'll need to extract it from the VPK)
You can set the color range to white and add a control point for the tint

ag3killstreak: tint doesn't look like it's going to work.

A tint is a color blend, unless if there is a way to get tint to replace the color completely.

ag3killstreak: 255 0 0 shows as pink.

ag3killstreak: and.. I broke it.

ag3killstreak: I'm not sure if you can even access the
	Vector m_flNormColorMin;
	Vector m_flNormColorMax;
	
variables with the pcf files.

src/game/server/swarm/asw_marine.cpp
ag3killstreak: Currently, g_rgbaStatsReportPlayerColors is only on the client.

If it's moved into shared then I can use it on the server code.

ag3killstreak: I moved it into a shared file so it can be accessed on server.

I also added code to set lasers to resource index if valid.

src/game/shared/particle_property.cpp
ag3killstreak: Yeah, I don't think control points are going to work.

The m_flNormColorMax and m_flNormColorMin variables don't appear to be included in the pcf data values.

Modifying the definition may be the only way we can product these color overrides.

If we had access to the particles files then we could set up something better but I'm not seeing any other solutions. Maybe you can find something better?

ag3killstreak: I may have found something. Tint won't work, but Remap Control Point to Vector might work.

I'm trying to program it in and see if it has desired result.

ag3killstreak: I can get the desired results in particle editor, but the setcontrolpoint method doesn't update the color while in actual game.

Not sure what I'm doing wrong.

ag3killstreak: ah, I get it now.

The Remap vector range was 0.0 to 1.0. I was treating it as a 0 - 255.

It is updating the color now. The result is not the same as m_flNormColorMax in definition but it is better than tint.

ag3killstreak: Ok, I uploaded the new test code.

ag3killstreak:
I'm in the process of converting the netprop into an encoded Int32. I figured using 12 bytes to store a color was just wasteful.

ag3killstreak:
:>

ag3killstreak:
Code is about ready to be checked again.

There is one issue that I can't figure out atm. The friendly fire effect does not show for the RGB effects. It however does show for classic forced mode. I tried changing the effect to one inside the pcf file but it made no difference.

ag3killstreak:
I fixed the FF particle effect. It turns out the pcf file bugged somehow so I had to delete the children and readd them.

PR #352, opened 2022-10-12
if there is no meelee attack, set zero

mithrand0:
Related to #14756184


BenLubar:
would 1 be a safer default?

PR #355, opened 2022-10-12
add additional check to make sure getPath() actually returns something

mithrand0:
related to #14756383


mithrand0:
not relevant

PR #364, opened 2022-10-19
detect clientside sprites and turn them into serverside ones. fixes #15

### LINKED ISSUE #15: env_sprite_clientside
if a "clientside" sprite which was meant to be serversided is detected, it should be destroyed and a a new serverside sprite with same properties should be created. instead of only reporting about the issue to the console

jhh8:


BenLubar:
do valid env_sprite_clientside still render? looks like this deletes quite a bit of code related to loading those

jhh8:
Fixed by orange in 7c29a07

PR #374, opened 2022-10-25
Deima Bug Fix

ag3killstreak:
I got stuck on bridge with bloodhound and we couldn't progress so I looked into it.

I believe it was caused by a marine skipping past the trigger.

I extended the trigger across the whole edge of the bridge so it should trigger each time. The reset trigger delay was dropped to 0.1 from 1.0 just in case that was also a factor.

### DIFF
--- a/contentsrc/mapsrc/ASI-Jac2-Deima.vmf
+++ b/contentsrc/mapsrc/ASI-Jac2-Deima.vmf
@@ -2,7 +2,7 @@ versioninfo
 {
 	"editorversion" "400"
 	"editorbuild" "7397"
-	"mapversion" "3300"
+	"mapversion" "3301"
 	"formatversion" "100"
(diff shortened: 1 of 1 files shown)


contentsrc/mapsrc/ASI-Jac2-Deima.vmf
```diff
@@ -164083,10 +164083,10 @@ entity
 {
 	"id" "1100771"
 	"classname" "trigger_multiple"
-	"origin" "8104 6872 -8"
+	"origin" "8472 6872 -8"
 	"spawnflags" "16"
 	"StartDisabled" "0"
-	"wait" "1"
+	"wait" "0.1"
```
BenLubar, line 164089: OnStartTouch isn't affected by the wait timer, so this should probably be reverted to 1 second

PR #390, opened 2022-11-08
suggestion: buff power fist first two attacks

jhh8:


mithrand0:
Why we buff it? What do others think of buffing an item?

Last time we changed things, the community had feedback after. Might be good to open a discussion item on Steam before merging in this change.

BenLubar:
@jhh8 should the 40 be changed? the old multipliers were 1.5/1.5/60 and the new ones are 3.5/3.5/140. Not sure if that part is intentional.

PR #393, opened 2022-11-09
env_shake: add a flag to force screen shake on clients with asw_camer…

jhh8:
…a_shake 0

### DIFF
--- a/src/game/client/view_effects.cpp
+++ b/src/game/client/view_effects.cpp
@@ -685,6 +685,7 @@ screenshake_t *CViewEffects::FindLongestShake()
 }
 
 ConVar asw_camera_shake( "asw_camera_shake", "1", FCVAR_NONE, "Enable camera shakes" );
+ConVar asw_camera_shake_forced( "asw_camera_shake_forced", "1", FCVAR_NONE, "Disables forced camera shakes. Forced camera shakes are used by few maps and perform a screen shake even when asw_camera_shake is set to 0." );
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


BenLubar:
asw_camera_shake is an accessibility option. not sure if it should be bypassable.

GmaOrange:
maybe second convar to disable also bypass or something like this, so users can disable only grenade launcher annoying things.

BenLubar:
Makes sense.

jhh8:
another example: research center map in amber project campaign, there is a water bridge section which makes the view blurry for players. that can be disabled with mat_postprocess 0 command and those people will have a slight advantage. i personally would prefer that an effect like that wouldnt be possible to turn off, but if you guys think it should be possible since it could be too annoying or some other reason then sure, i could add another cvar for shake which would also disable the forceful shakes

src/game/client/view_effects.cpp
```diff
@@ -685,6 +685,7 @@ screenshake_t *CViewEffects::FindLongestShake()
 }
 
 ConVar asw_camera_shake( "asw_camera_shake", "1", FCVAR_NONE, "Enable camera shakes" );
+ConVar asw_camera_shake_forced( "asw_camera_shake_forced", "1", FCVAR_NONE, "Disables forced camera shakes. Forced camera shakes are used by few maps and perform a screen shake even when asw_camera_shake is set to 0." );
```
BenLubar, line 688: Looks like this *En*ables shake, not *Dis*ables.

PR #397, opened 2022-11-10
asw_door: make it possible to set custom door health and dent percentage

jhh8:


BenLubar:
What's the use case for this? I'd prefer to keep it how it was for gameplay consistency unless there's a specific reason to do this.

jhh8:
More variety for mappers

jhh8:
for example a mapper can make a fence prop a door (in this case it would serve like a prop which players would have to break to unblock their way or something, and it would have a healthbar) but since fences are weak the mappeer could do 400 health instead of 1800. Or if a mapper wants a regular door but with a dent, he is currently limited to either 100% health, 66% health or 33% health, maybe someone would want 50% or 20% or some other value

BenLubar:
Sounds reasonable.

PR #401, opened 2022-11-14
particle crash

mithrand0:
Possible crash in particles when using multiple threads

https://partner.steamgames.com/errors/errorid/14879711


BenLubar:
Won't this kill particle rendering for the non-threaded render path?

mithrand0:
> Won't this kill particle rendering for the non-threaded render path?

Should be fixed

Issue #356, opened 2022-10-12
high cpu load on Lana's Bridge


mithrand0:
The cpu load on Lana's Bridge is extremely high compared to other maps, often hitting 100% on higher difficulties with full lobbies. It would be good to check if this map can be optimized.

jhh8:
did you check on beta? i did some changes in #334

jhh8:
still wondering if on beta this is fine or need mmore changes

mithrand0:
yes, still the same fps drops to single digits due very high cpu load

wesaamt:
Hi

PR #425, opened 2022-12-01
lana1: optimizations

jhh8:
replaced a lot of asw_drone_jumper spawners with normal asw_drone, it looks like pathfinding calculations are heavy with asw_drone_jumper because the map's nodegraph for them is connected from the place they spawn (below the bridge in snowy area or near the door) all the way to the spawn, and with normal asw_drone the nodegraph is isolated into two pieces before the fall and after the fall, just like for marines.

Also the nodegraph itself was unnecesarily complex, removed half of the nodes. cut down nodegraph build time from 20 seconds to 10 seconds.

All in all the map should be a bit easier not just in terms of lack of lag but i think its worth it


mithrand0:
related to issue #356

BenLubar:
Can we use asw_simple_drone for the "cutscene" aliens?

PR #431, opened 2022-12-09
vehicles: marine driving and riding activities

jhh8:


BenLubar:
If we can get a solid way of adding animations to marines, that would be a huge boon.

PR #434, opened 2022-12-17
acc32: add new airlocks and something interesting on map6

jhh8:


BenLubar:
Does acc_complex need any updates?

Issue #437, opened 2022-12-19
Change to custom panel message for DNA sample in BioGen Corporation (rd-bio3biogenlabs)


Blueberryy:
Should be changed to something like "Collect DNA" or something.
![reactivedrop_u7hLXdjxPv](https://user-images.githubusercontent.com/36592509/208401400-564f0f5d-21c1-475f-8fba-3c26d27bf995.png)

BenLubar:
What other verbs might we want to replace "Use Panel" with? So far I can think of at least "Repair Circuit" for maps where the wire hack isn't a hack.

PR #438, opened 2022-12-19
make it possible to customize button area icon texts, see #437

jhh8:
use panel message,
need tech message,
exit panel message,
hack panel message

### DIFF
--- a/src/game/client/swarm/c_asw_button_area.h
+++ b/src/game/client/swarm/c_asw_button_area.h
@@ -44,11 +44,6 @@ public:
 	CNetworkVar(float, m_fHackProgress);
 	CNetworkVar(float, m_flHoldTime);
 	
-	virtual const char* GetNoPowerMessage() { return m_NoPowerMessage; }
-	virtual const char* GetUseIconMessage() { return m_UsePanelMessage; }
-	virtual const char* GetLockedIconMessage() { return m_NeedTechMessage; }
-	virtual const char* GetExitPanelMessage() { return m_ExitPanelMessage; }
-	virtual const char* GetHackPanelMessage() { return m_HackPanelMessage; }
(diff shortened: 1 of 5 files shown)


src/game/client/swarm/c_asw_button_area.h
```diff
@@ -45,7 +45,17 @@ class C_ASW_Button_Area : public C_ASW_Use_Area
 	CNetworkVar(float, m_flHoldTime);
 	
 	virtual const char* GetNoPowerMessage() { return m_NoPowerMessage; }
-	char		m_NoPowerMessage[255];	
+	virtual const char* GetUseIconMessage() { return m_UsePanelMessage; }
```
BenLubar, line 48: Remove these virtual methods or move them to the cpp file and have them handle the default value automatically.

src/game/server/swarm/asw_button_area.cpp
```diff
@@ -615,6 +621,26 @@ bool CASW_Button_Area::KeyValue( const char *szKeyName, const char *szValue )
 		Q_strncpy( m_NoPowerMessage.GetForModify(), szValue, 255 );
 		return true;
 	}	
+	else if ( FStrEq( szKeyName, "usepanelmessage" ) )
```
BenLubar, line 624: For safety (because many existing maps won't have these keyvalues set) we should do `m_*.GetForModify()[0] = '\0';` in the constructor.

src/game/client/swarm/c_asw_button_area.h
jhh8: What do you mean by handle the default value automatically?

BenLubar: Basically, the functions should look like `GetUseIconText` and not like `GetUseIconMessage`

PR #474, opened 2023-01-21
add flamerboost convar, resolves #473

### LINKED ISSUE #473: [request] flammer boost fix only applies to ranked server
I am a speed runner. I understand that the ranked servers are places for battle skill competition. Banning all kinds of bugs and skips on ranked servers are reasonable. But to us speed runners, We treat bugs and skips as another kind of skills that need to learn and practice So that we can get a perfect flammer boost or something.

flammer boost is an old bug/glitch. Why not leave it as a game feature for non-ranked servers and lobbies. Some flammer boost spots need up to5 players' coop to save time. We have practiced for months to get a perfect run using flammer boost technique. The fix ruins all of our team's efforts. Just because some players are using it in ranked servers. 

To be honest, speedrun.com needs to upload video proof, those skips. Players will know these skips eventually. Those speedrun skills are not intend for ranked servers. So devs could just tell everyone that bugs, skips and mods are forbidden in ranked servers and one will get banned If He uses these "techniques".

Please leave some spaces for speed runners.

jhh8:


BenLubar:
I don't think we should make a gameplay-affecting convar that's not a cheat, and I'm not confident in my ability to make it work properly if it has a different default for ranked servers.

The way I'd solve this is to add FCVAR_CHEAT, make the default 0, and have a workshop challenge that isn't rated in HoIAF that uses that convar.

ywgATustcbbs:
> I don't think we should make a gameplay-affecting convar that's not a cheat

Actuality, this is a game bug or a game feature in a player's view for years until being utilized to farming. The bug itsself is ok most of the time except for ranked servers.

As far as i know,  only a small part of the players plays ranked games, most of which are hard core players knowing what  can do. The bug only comes into devs eyes only because a few players' farming. If flammer boost is marked as a cheat, it is actually banned for speedrunning anyway. Cheat cvars are not allowed to be changed in a formal speedrun.

 Btw, It's a lot fun to use flammer boost in a non-competitive lobby or server. More over, there are cvars that affects game play while not marked ss cheating. Camera rotation, death cam, Camera shake, cmdrate, lerp ratio etc.

jhh8:
i think its fine to not keep it as a cheat, its like saying we are allowing it to be used but if someone wants they can disable it

BenLubar:
We had a previous situation with a similar convar to the one being added in this PR: `asw_marine_rolls`.

The idea I have here is to do the same thing with this one: mark it as FCVAR_CHEAT, and therefore allow it to be used in challenges. Most custom challenges change at least one cheat convar, so this would be no different than any other setting challenges can change.

That way, speedruns could still be done with this, as an "extinguisher boost" category, just like how Half-Life 2 has a "technically not cheating" category where commands like ch_createairboat are allowed.

mithrand0:
> We had a previous situation with a similar convar to the one being added in this PR: `asw_marine_rolls`.
> 
> The idea I have here is to do the same thing with this one: mark it as FCVAR_CHEAT, and therefore allow it to be used in challenges. Most custom challenges change at least one cheat convar, so this would be no different than any other setting challenges can change.
> 
> That way, speedruns could still be done with this, as an "extinguisher boost" category, just like how Half-Life 2 has a "technically not cheating" category where commands like ch_createairboat are allowed.

Agree with Ben on this one.

ywgATustcbbs:
> Challenge modes can change convars regardless of whether they have FCVAR_CHEAT set and regardless of whether sv_cheats is set.

I mean, that, if i don't want any challenge And just play on normal or other difficulties, Then i have to set sv_cheat 1 to before i can enable the boost. The record will be rejected by lb, am i right?

BenLubar:
Yes, leaderboard entries are not uploaded if `sv_cheats 1` was set during the session. The engine won't let you set `sv_cheats 1` in a lobby anyway.

ywgATustcbbs:
> Yes, leaderboard entries are not uploaded if `sv_cheats 1` was set during the session. The engine won't let you set `sv_cheats 1` in a lobby anyway.

I see. It is actually banned in speedruns for leaderboard.

ywgATustcbbs:
> Yes, leaderboard entries are not uploaded if `sv_cheats 1` was set during the session. The engine won't let you set `sv_cheats 1` in a lobby anyway.

Since the bug exists for years and doesn't really influence game experience except for  ranked servers, i suggest leaving it as is and don't mark it as cheating for non-ranked casual games. Just ban it on ranked servers is enough. 

For speedrun leaderbosrd, the bug will increase a lot of difficulties of practicing and requires teamwork for most of the maps (which are the aims of speedruns, too), i think it is a good glitch to encourage team practices. Instead, cheating records are much more harmful to leaderboard and speedrun community.

All my concerns, opinions and suggestion are presented. Anyway, it is the devs to make the final decision.

mithrand0:
> > Yes, leaderboard entries are not uploaded if `sv_cheats 1` was set during the session. The engine won't let you set `sv_cheats 1` in a lobby anyway.
> 
> Since the bug exists for years and doesn't really influence game experience except for ranked servers, i suggest leaving it as is and don't mark it as cheating for non-ranked casual games. Just ban it on ranked servers is enough.
> 
> For speedrun leaderbosrd, the bug will increase a lot of difficulties of practicing and requires teamwork for most of the maps (which are the aims of speedruns, too), i think it is a good glitch to encourage team practices. Instead, cheating records are much more harmful to leaderboard and speedrun community.
> 
> All my concerns, opinions and suggestions are presented. Anyway, it is the devs to make the final decision.

If I compile a custom client now, I can upload any score with any time for any map. As long as leaderboards aren't protected and verified, they will never useful for comparison. 

Apart from that, the speedrun community has one rule: **video proof or it didn't happen**. The reason for this, is because speedruns are about skill, and not exploiting unknown methods to gain advantage. We wouldn't be able to rule TAS runs, or lag exploits for example.

ywgATustcbbs:
> If I compile a custom client now, I can upload any score with any time for any map. As long as leaderboards aren't protected and verified, they will never be useful for official speedruns.

Yes, one team can cheat. But not all teams will cheat. Just like the ranked servers have been running with this bug for more than half a year and nothing happened until a few guys abused it. 

> Apart from that, the speedrun community has one rule: **video proof or it didn't happen**. The reason for this, is because speedruns are about skill, and not exploiting unknown methods to gain advantage. We wouldn't be able to rule TAS runs, or lag exploits, or other methods like cheatengine.

speedrun.com requires video proof. Ingeneral, a challenge is forbidden due to it is a modification to the game. Sinece sv_cheat can't be set in a lobby, one can't srt the boost. This means the boost is actually banned.

mithrand0:
@BenLubar 

I propose this as solution for this discussion:
https://github.com/ReactiveDrop/reactivedrop_public_src/issues/477

The idea is based on the "half life technically not cheating category".

Issue #479, opened 2023-01-26
Speedrun.com integration


ywgATustcbbs:
Come up with some new possible solutions. Sorry to post them here because i use a phone to edit texts and steam app is unfriendly to long text editing.

I see the current possible solution is to use speedrun.com, But few ssrd players knows the website. It needs recognition and promotion from the devs before it is accepted by the Asrd players. Technically, it can replace or integrate with the game to encourage players to perform runs that can be verified, and discourage/expose those cheaters, further maintaining a good community environment.

There are several solutions to integrate speedrun.com into lb. Followings are sorted based on the works needed.

1st solution. No need to change game logic. But to retrive info from speedrun.com and display it in game.
1. Split the current in-game lb screen into two parts, display verified speedrun.com results as verified on top half, and display the current lb as un-verified on the bottom half.
2. Done. Thus will encourage players to submit records for verification.

2nd solution is based on the 1st one:
1. contact moderators on speedrun.com, increase/expand speedrun categories for each map and difficulties. Sure, the moderator will have a lot of works to do. But i think community will get it done.
2. Retrieve data from speedrun.com, display them.
3. Done.

3rd solution (based on previous one):
1. invite some moderators to join devs team as long-term community menbers. 
2. Review And revise the current speedrun rules to force using speedrun challenge as proposed 
3. Recruit more community moderators.
4. Done.

4th solution:
1. Do as speedrun.com, Add a speedrun submitting page to the official website. User fills in their steam I'd and other game info, post a YouTube link.
2. Find some community moderators to verify the submitted runs.
3. Display them in the game and on the official website.
4. This need much more web coding.

5th solution. Integrate with autorecoding.
1. Send each demo md5 and sha256 strings to website to protect records from demo editing.
2. Community moderators verify top records,  websit auto verify other records using demo file checks.
3. No video and demo verification is required, unless it breaks the top 20 and 
4. Top 100 records.
5. Players who have broken top records will have several days or weeks to prepare video and/demo using local demo files.
6. Once demo uploaded to the website, check md5 in the database.
7. Once passed, run auto verification. And if needed, perform community verification.
8. Only permanently keeping top 20 or 50  demos for further reporting.

These solutions don't need much website resources. Perhaps, devs can try to contact/regrouping community members. I think it worth trying.

mithrand0:
Integration would mean, someone needs to code it, it can be a lot of work.

ywgATustcbbs:
[...]

PR #488, opened 2023-02-02
speedrun leaderboard enhancement demo

ywgATustcbbs:
This is a demo for leaderboard enhancement. It mainly hooks the source at to places: 1.right after the updateleaderboardscore function call. 2. In the renaming demo function.

The demo Currently has build errors in debug. But It can be built in release. The lnk error is really beyon my abilities.

Please evaluate it. If permitted, i can finish the rest of the displaying part.


ywgATustcbbs:
Made a mistake. Should have used `AttachLeaderboardUGC()` Instead of
`UploadLeaderboardScore()`. And call
 `GetDownloadedLeaderboardEntry()` after `DownloadLeaderboardEntries()` . In This way, the signature string will not influence the current lb functionality at all. Will fix it later.

mithrand0:
The PR approach is not usable.

Regarding this functionality, after talking, the issue #479 was closed. We added some forum information, and information to the site, to satisfy the needs for the members who want to utilize speedrun.com.

Closing this PR in favor of the actions mentioned above.

PR #505, opened 2023-02-14
simple hacking: hacking scale speeds up download

jhh8:
currently we cant make a hack be fast from altering hacking scale because how big of an impact download time has, which is not good, so make hacking scale speed up the download too


BenLubar:
I'd prefer a separate convar that scales download times for both bots and players.

jhh8:
having it in one convar seems more compact and intuitive for me, challenge makers wont know to change both convars to achieve what they want, also would keep some challenges how they worked before, singleplayer challenge on beta downloads very slowly the computer hack now, which is a big deal because many speedruns were done there

BenLubar:
The reason I'd prefer separate convars is that this affects data download times for the NPC style of hacking but not the player character style. And I'd like to avoid having one convar that does two different things if at all possible.

PR #511, opened 2023-02-25
death type - sentry explosion

jhh8:


BenLubar:
New ones need to be inserted at the end of the list.

jhh8:
I thought so at first too, but it looks like it only matters for the death cause to be in same position in the string array as in the enum. I have tested it too and everything worked

BenLubar:
Basically, the problem with putting things in the middle of the list is that if the client and the server disagree about the order of the list, the client is going to add to the wrong stat.

PR #561, opened 2023-04-06
bossbar: radius keyvalue

jhh8:


src/game/client/swarm/c_rd_boss_bar.cpp
```diff
@@ -34,6 +68,8 @@ void C_RD_Boss_Bar::PostDataUpdate( DataUpdateType_t updateType )
 	{
 		m_flBarValuePrev = m_flBarValue;
 		m_flBarValueLastChanged = gpGlobals->curtime - flSustain - flInterpolate;
+
+		SetNextClientThink( gpGlobals->curtime + 0.1f );
```
BenLubar, line 72: use `CLIENT_THINK_ALWAYS` and remove the other instance of SetNextClientThink

src/game/client/swarm/c_rd_boss_bar.h
```diff
@@ -24,10 +24,15 @@ class C_RD_Boss_Bar : public C_BaseEntity
 	float m_flBarFlashSustain;
 	float m_flBarFlashInterpolate;
 	bool m_bEnabled;
+	float m_flBarRadius;
+	bool m_bBarTooFarAway = false;
```
BenLubar, line 28: We usually initialize member variables in the constructor; move the = false there for consistency.

src/game/client/swarm/c_rd_boss_bar.cpp
jhh8: didnt want it to be too expensive

BenLubar: Your ClientThink is cheap enough that it can run every frame without anyone noticing.

Re-scheduling might cost more than the rest of the ClientThink.

PR #565, opened 2023-04-09
acc32 tweaks

jhh8:
map1 to map4
added door frame models

map2
made upper area right before reactor room into a supply area made first breifing camera not show the shower alien which falls

map5
added bigger shower particles and steam particles when biomass extinguishes

map6
added a pipeleak at the sample at top right, so zombine spawns are less rare added a ladder closer to the muon gem fight
made boss bars have a radius


BenLubar:
Did you fix the music cue you said you wanted to?

PR #574, opened 2023-04-25
optimisation, calculation results are unused

jhh8:
called every frame

compiler should be optimising that but doesnt look like it does


BenLubar:
These values are used.

jhh8:
Ok

PR #604, opened 2023-06-04
make rifle grenade explode rather than disappear, add a convar for li…

jhh8:
…fetime


BenLubar:
It's very weird that it was set up like this.

PR #607, opened 2023-06-08
add chat rate limit

qq578023708:
allow set player chat rate limit

### DIFF
--- a/src/game/server/player.h
+++ b/src/game/server/player.h
@@ -1,4 +1,4 @@
-//===== Copyright © 1996-2009, Valve Corporation, All rights reserved. ======//
+//===== Copyright ?1996-2009, Valve Corporation, All rights reserved. ======//
 //
 // Purpose: 
 //
@@ -110,6 +110,12 @@ enum PlayerPhysFlag_e
 
 	// If you add another flag here check that you aren't 
 	// overwriting phys flags in the HL2 of TF2 player classes
+};
+
+enum PlayerSpeakResult
+{
+	SPEAK_RESULT_OK,
+	SPEAK_RESULT_RateLimited
 };
 
 //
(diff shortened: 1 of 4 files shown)


src/game/server/player.h
```diff
@@ -622,6 +628,7 @@ class CBasePlayer : public CBaseCombatCharacter
 	// say/sayteam allowed?
 	virtual bool		CanHearAndReadChatFrom( CBasePlayer *pPlayer ) { return true; }
 	virtual bool		CanSpeak( void ) { return true; }
+	virtual PlayerSpeakResult        CheckSpeak( void );
```
BenLubar, line 631: Unless we're planning on making this a lot more complicated very soon, I'd prefer that this just returns a bool.

src/game/shared/swarm/asw_gamerules.cpp
```diff
@@ -204,6 +204,7 @@ extern ConVar old_radius_damage;
 	ConVar rd_mapcycle_endless( "rd_mapcycle_endless", "0", FCVAR_ARCHIVE, "Automatically select the next Endless mission." );
 	ConVar rd_mapcycle_bonus( "rd_mapcycle_bonus", "1", FCVAR_ARCHIVE, "Automatically select the next Bonus mission." );
 	ConVar rd_mapcycle_ignore( "rd_mapcycle_ignore", "", FCVAR_ARCHIVE, "Comma-separated list of map filenames (no .bsp) that cannot be selected by map cycle." );
+	ConVar rd_chat_ratelimit ( "rd_chat_ratelimit", "1", FCVAR_NONE, "Set player chat rate,default 1 second" );
```
BenLubar, line 207: Don't put the default value in the description, the engine will automatically do that so it'll end up there twice.

jhh8:
default 0? there already is some sort of limiting, not sure if clientside or serverside

src/game/server/player.h
jhh8: should just edit the canspeak function instead of a new one

PR #616, opened 2023-07-16
misc pedantic fixes (+ comments for modders)

kablooey123:


src/game/client/glow_overlay.cpp
```diff
@@ -119,7 +119,7 @@ bool CGlowOverlay::Update()
 	return true;
 }
 
-ConVar building_cubemaps( "building_cubemaps", "0" );
+ConVar building_cubemaps( "building_cubemaps", "0", FCVAR_CHEAT | FCVAR_HIDDEN );
```
BenLubar, line 122: I would use FCVAR_DEVELOPMENTONLY here instead.

PR #622, opened 2023-07-28
Up default projected texture limit and add Flashlight settings under "Visual Effects".

goodusername123:
- Up max projected texture limit from 1 to 10.
- Remove ARCHIVE flag from "rd_max_depth_texture_shadows" CVAR since it causes issues due to config.cfg being executed too late.
- Add Advanced setting toggles under "Visual Effects" for Flashlight projected texture shadows and Flashlight dlight creation.

My rationale for upping the projected texture limit to such a seemly large number is because when messing around with "rd_flashlightshadows" in a full 8 slot game with everyone having flashlights I noticed that even with 9 or 10 projected textures on screen at once the game ran beautifully at 4K with max in game settings and had no visual artifacts even while action was happening so I see the current limit of "1" as pretty artificial especially since a lot of other source engine games/mods have this limit changed with little to no issues. as for why I chose "10" specifically it's because it is the minimum needed for a worst case scenario of 8 marines all with flashlights plus the current map's on-screen projected texture plus an extra one as a backup and for evening out the total number.

my suggestion for what the new strings should say are as follows although these can be worded to however is seen fit:
- RD_AdvancedSettings_VisualEffects_FlashlightShadows = Flashlight Shadows
- RD_AdvancedSettings_VisualEffects_FlashlightAmbientLighting = Flashlight Ambient Lighting

The reasons I am submitting this as a draft pull request is because:
1. I am uncertain if this should be for the public or beta branch as I couldn't find any proper information as to what kinda changes should go to which branch (also probably doesn't help that I don't know how to use github that well) although I assume the weight of my changes should be fine to push out to the next public release.
2. This pull request can't be immediately merged in as a commit is needed on the translations repo for the new strings (hopefully someone is able to do this for me as I don't know how to set that up).
3. I wanted to have the "Flashlight Shadows" setting not display unless "Dynamic Lighting" is enabled since "Dynamic Lighting" is supposed to turn off projected textures however it seems to only affect map projected textures and not ones created by the flashlights, although this is a small detail so it might be fine if no easy proper solution is available/found.
4. I want a little bit of input as maybe there is something I didn't consider however I feel pretty good about these current changes for the most part.

maybe I am misusing or misunderstanding the draft pull request feature of github? but it shouldn't be too much of an problem as I should be able to just "mark as ready for review" and have it turn into a normal pull request hopefully.

### DIFF
--- a/reactivedrop/resource/rd_advanced_settings.txt
+++ b/reactivedrop/resource/rd_advanced_settings.txt
@@ -521,6 +521,17 @@
 		}
 		TOGGLE
 		{
+			title "#RD_AdvancedSettings_VisualEffects_FlashlightShadows"
+			cvar "rd_flashlightshadows"
+			hide_unless "rd_env_projectedtexture_enabled"
+		}
+		TOGGLE
+		{
+			title "#RD_AdvancedSettings_VisualEffects_FlashlightAmbientLighting"
+			cvar "rd_flashlight_dlight_enable"
+		}
+		TOGGLE
+		{
 			title "#RD_AdvancedSettings_VisualEffects_OtherEffects"
 			cvar "rd_info_particle_system_enabled"
 		}
--- a/reactivedrop/resource/rd_advanced_settings.txt
+++ b/reactivedrop/resource/rd_advanced_settings.txt
@@ -523,7 +523,6 @@
 		{
 			title "#RD_AdvancedSettings_VisualEffects_FlashlightShadows"
 			cvar "rd_flashlightshadows"
-			hide_unless "rd_env_projectedtexture_enabled"
 		}
 		TOGGLE
 		{
--- a/src/game/client/clientshadowmgr.cpp
+++ b/src/game/client/clientshadowmgr.cpp
@@ -1,4 +1,4 @@
-//===== Copyright © 1996-2005, Valve Corporation, All rights reserved. ======//
+//===== Copyright Â© 1996-2005, Valve Corporation, All rights reserved. ======//
 //
 // Purpose: 
 //
@@ -1409,7 +1409,7 @@ void CClientShadowMgr::CalculateRenderTargetsAndSizes( void )
 		m_nDepthTextureResolution = atoi( CommandLine()->ParmValue( "-sfm_shadowmapres", defaultRes ) );
 	}
 	extern ConVar rd_max_depth_texture_shadows;
-	m_nMaxDepthTextureShadows = bTools ? MAX_DEPTH_TEXTURE_SHADOWS_TOOLS : CommandLine()->ParmValue( "-rd_max_depth_texture_shadows", MAX_DEPTH_TEXTURE_SHADOWS );	// Just one shadow depth texture in games, more in tools
+	m_nMaxDepthTextureShadows = bTools ? MAX_DEPTH_TEXTURE_SHADOWS_TOOLS : CommandLine()->ParmValue( "-rd_max_depth_texture_shadows", MAX_DEPTH_TEXTURE_SHADOWS );	// Limit shadow depth texture amount in games, more in tools
 }
 //-----------------------------------------------------------------------------
 // Constructor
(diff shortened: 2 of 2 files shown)


BenLubar:
Depth textures are (by default) 1024x1024, and the format varies by GPU but I think we can assume it's 4 bytes per pixel or less (although some formats like RGBA32323232F are 16 bytes (128 bits!) per pixel.) So 4 megabytes of VRAM per depth texture.

I've got hardware statistics for AS:RD players who participated in the June 2023 Steam Hardware Survey; here's VRAM:

image

If we're increasing the default depth texture count from 1 to 10, that adds 9*4MiB, or 36MiB of depth textures. For some of the lowest-end machines (rounded to 512 MiB of VRAM), that might make them unable to load certain maps.

However, it would be nice to make the ability to change shadow quality settings easier. Right now, from what I can see in the code, it looks like the convar is completely non-functional and it only looks at the command line argument.

BenLubar:
Related: f44efcfb9974bcf2ba5223906044502b15c16b6c

goodusername123:
Whoops I feel very dumb right now because I have should have checked the beta branch before making this pull request as I now see that the UI part of this request is completely invalid as the beta branch has equivalent options to the ones I have added which means that the new strings I was attempting add are just a waste of translation efforts from what it seems.


@BenLubar Yeah it would be nice to have a in-game way of changing the projected texture limit as the current value of `1` makes the game end up fighting over what gets to use the single projected texture when `rd_flashlightshadows` is enabled, I think for the time being maybe at least increasing the default to a more reasonable value such as `2` or `3` instead of my originally proposed `10` would be worthwhile so the game would only start fighting over projected textures when more then one player has a flashlight, Also if/when a in-game setting was added for changing the max projected texture count then it could probably be simplified to options like "Low, Medium, and High" for example since displaying just raw numbers could be seen as pretty un-friendly way of exposing the option to players.

BenLubar:
Don't feel too silly - I made that commit as a result of your PR!

BenLubar:
@geotavros how do you feel about forcing the shadow buffer count to be at least [value] if rd_flashlightshadows and projected textures are both enabled?

geotavros:
I'd look how it is done on other(possibly Valve) games. Afaik most Valve games have 1 shadow buffer. If this setting can significantly degrade rendering performance it needs to be stated so in the UI. So that players understand what they are changing and what it affects. Ideally it'd be good to somehow educate users of what setting causes what result.

BenLubar:
Basically, each depth texture costs a certain amount of VRAM (if you don't have enough, it'll just error out while you're trying to load a level) and also requires drawing the scene from multiple angles. If you have a full squad with flashlights, a mapper-placed projected texture, and some water in the scene, that requires rendering the scene 13+ times per frame from different angles.

The depth render isn't as expensive as a full render with antialiasing and complex shaders and all that, but it does need to compute where all the polygons are and which one is covering which pixel. And for alpha-tested materials, it does need to do some texture lookups during that.

mithrand0:
This PR is quite old, closing it.

geotavros:
Having a UI option to set the number of projected textures would be nice. I, myself, was always using a startup argument `-rd_max_depth_texture_shadows 8`, but because my hardware allows this. I know that a lot of players disable projected textures to get higher FPS. Also a lot of players play on laptops and slower hardware.

Also, allowing mappers to set the resolution of `env_projectedtexture `and `env_global_light `would be cool too. 

> Depth textures are (by default) 1024x1024

AFAIR(last time I looked a few years ago) it depended on client's game resolution. Depth texture would be a square power of 2 but less than game client resolution. So if your game resolution is less than 1920x1080 the depth texture would be 512x512.

Also the important thing is the shadow filter size(`m_flShadowFilterSize`), which needs to be lowered for higher resolution depth texture. Otherwise there's no point in higher resolution as it will be blurred by the filter.

I've been working on a map utilizing `env_global_light`, but for nice looking shadows it required at least 2048x2048 depth texture. Here's a screenshot of 4096 depth texture and 0.2 `m_flShadowFilterSize`
https://steamcommunity.com/sharedfiles/filedetails/?id=2185405119

Issue #624, opened 2023-08-02
Game folder server locking


mithrand0:
Some users (try) run multiple instances from the same game installation folder. This results in unreliable behavior, starting with workshop items not being downloaded, partially downloaded, or other interference.

We should lock the game folder with `lock_exclusive` at start, allowing only one instance at a time from it.

jhh8:
I disagree, there are more positives about allowing that. You dont need to install multiple copies of the game to run multiple servers (many saved gigabytes) its also possible to implement cross-server chatting with vscript. Issue with workshop is solvable if you run one server first and let it install everything, then launch the other servers

mithrand0:
You can disagree, but Steam SDK states the files are not to be touched:

https://partner.steamgames.com/doc/api/ISteamUGC:

> Do not access the workshop item on disk until the Callback [DownloadItemResult_t](https://partner.steamgames.com/doc/api/ISteamUGC#DownloadItemResult_t) is called.

This has to do with file locking of Steam UGC content. 

If want to support multiple instances, make it official. Right now it causes issues.

> Issue with workshop is solvable if you run one server first and let it install everything, then launch the other servers

Sounds not very workable.. please share the script you use on your servers for this.

jhh8:
so what you're saying is when two instances are running, one tries to update and another one is accessing it, the addon files get corrupted, set to read only forever or something and server is not able to update them ever again, even if owner starts running only one instance like a normal server?

mithrand0:
yes, I have seen that case.

it is hard to reproduce, because of the exact timing, but when it happens, you need to throw away the workshop and steam ugc folders. That is why I decided to go separate folders in the past, solves a lot of issues.

mithrand0:
Sourcemod sort of has a similar issue when sharing it's data: https://wiki.alliedmods.net/Multiple_or_Forked_Servers_(SourceMod)

Seems Left 4 Dead 2 implemented a `-fork <num>`  parameter to support multiple servers from the same game folder.

BenLubar:
Might be able to just set each server instance to use a separate workshop folder

mithrand0:
Keep in mind, if the folder is shared, `steamcmd` can't be reliable used to update the game between server restarts. Also, if workshop has separate folders, the benefit of a shared install become smaller. 

If space is the issue, better look into ways to remove textures and sounds from the server install, theoretically those shouldn't be needed on a server and I can imagine they might eat up a considerable amount of size.

[...]

PR #625, opened 2023-08-02
file locking

mithrand0:
implements #624, prevents workshop issues with file already locked, and UGC content that gets stale.

@BenLubar up to you to merge it or not, for discussion, see ticket above.


jhh8:
what about also a convar to not do this

BenLubar:
Won't this cause problems when the level changes?

mithrand0:
It won't, once the file is locked it doesn't check anymore.

PR #626, opened 2023-08-02
new map voting system

jhh8:
now all voters count towards the vote fraction instead of all players on the server.

in the previous system not voting is treated as voting no.

new system will not factor in afk people and people who don't care about next map and want to leave it for others to vote

### DIFF
--- a/src/game/client/swarm/c_asw_player.h
+++ b/src/game/client/swarm/c_asw_player.h
@@ -127,6 +127,8 @@ public:
 	virtual int GetHealth() const;
 	virtual int GetMaxHealth() const;
 
+	bool IsConnected() { return true; }
+
 // Called by shared code.
 public:
 	void ItemPostFrame();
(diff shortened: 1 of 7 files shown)


BenLubar:
We need some way of limiting how often a map vote can be called (per player) or this will be used to spam votes.

jhh8:
how so? why isnt it an issue with current system?

jhh8:
we could make it that if a vote was started in the middle of the mission, then it works like the previous system. And i think lowering asw_vote_duration would help too

BenLubar:
> how so? why isnt it an issue with current system?

With the current system, if someone is spamming map votes, not voting is the same as voting no, so the vote wouldn't go through.

> we could make it that if a vote was started in the middle of the mission, then it works like the previous system.

This could work, or we could add a per-player timeout for calling a map vote after calling a map vote

PR #637, opened 2023-08-17
memory leak detection

mithrand0:


BenLubar:
sv_memlimit already exists

mithrand0:
> sv_memlimit already exists

`sv_memlimit` is marked cheat, and I think does not notify players.

BenLubar:
Looks like it only works on Linux builds of SRCDS as well, ugh.

PR #649, opened 2023-09-22
Bugfix/sv memlimit exit

mithrand0:
Send `quit` multiple times, it (randomly) isn't received correctly the first time.

```Memory usage: 214 MiB (224481280 bytes); limit is 215 MiB
---- Host_NewGame ----
Host_NewGame on map dm_deima
Skill level changed 2
Skill level changed 2
Set Gravity 800.0 (0.250 tolerance)
Loading global spawn selection data
No map-specific spawn selection data for dm_deima
Cannot access ISteamInventory!
Level has follow hints 1
Created class baseline: 38 classes, 6655 bytes.
Connection to Steam servers successful.
exec: couldn't exec workshop.cfg
   VAC secure mode is activated.
Memory usage: 220 MiB (231481344 bytes); limit is 215 MiB
Memory limit exceeded. Requesting exit.
Requesting exit due memory limit exceeded. <<<
---- Host_NewGame ----
Host_NewGame on map dm_deima
Skill level changed 2
Skill level changed 2
Set Gravity 800.0 (0.250 tolerance)
Loading global spawn selection data
No map-specific spawn selection data for dm_deima
Cannot access ISteamInventory!
Level has follow hints 1
Created class baseline: 38 classes, 6655 bytes.
Requesting exit due memory limit exceeded. <<<
Memory usage: 226 MiB (237006848 bytes); limit is 215 MiB
Memory limit exceeded. Requesting exit.
Requesting exit due memory limit exceeded. <<<
Datacenter::EnableUpdate( 0 ), current state = 0
Datacenter::RequestStop, time 16.05, state 0
```

### DIFF
--- a/src/game/server/swarm/rd_memlimit.cpp
+++ b/src/game/server/swarm/rd_memlimit.cpp
@@ -13,12 +13,12 @@ inline unsigned int MemoryExceededShutdown( void* pParam )
 
 	// during a level change, the server doesn't always respond
 	// repeat this command
-	for ( int i = 0; i <= 10; i++ ) 
+	for ( int i = 0; i <= 32; i++ ) 
 	{
-		ConMsg("Requesting exit due memory limit exceeded.\n", delay);
-		engine->ServerCommand("quit\n");
+		ConMsg( "Requesting exit due memory limit exceeded.\n", delay );
+		engine->ServerCommand( "quit\n" );
 
-		ThreadSleep(delay);
+		ThreadSleep( delay );
 	}
 
 	return 0;
--- a/src/game/server/swarm/rd_memlimit.cpp
+++ b/src/game/server/swarm/rd_memlimit.cpp
@@ -9,13 +9,17 @@
 
 inline unsigned int MemoryExceededShutdown( void* pParam )
 {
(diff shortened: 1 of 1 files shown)


BenLubar:
Can we just call the C exit function?

mithrand0:
> Can we just call the C exit function?

With `quit`, players get disconnected with the message: `Disconnected, server is shutting down`. With a C exit, players would timeout without message.

BenLubar:
Hmm, if the problem is that we're not calling the quit concommand because the command queue is getting dropped, could we execute the command directly instead? Similar to how the vscript ConVars API lets you ignore FCVAR_CHEAT.

mithrand0:
> Hmm, if the problem is that we're not calling the quit concommand because the command queue is getting dropped, could we execute the command directly instead? Similar to how the vscript ConVars API lets you ignore FCVAR_CHEAT.

I would not know how.

Right now, I prefer a working fix, even if it's ugly. We can't keep restarting servers manually when they run out of memory.

BenLubar:
Fixed in ac5e143b459e782cfc46d9904c89ea3f776249ac

PR #652, opened 2023-10-07
cl_loadout_random command

jhh8:
this command selects random regular and extra weapons for you


BenLubar:
It'd be better if this was client-side and just sent a `cl_loadouta` command because then we could use it in the offline loadout editor and it'd be able to pick up weapons that have items associated with them in the future

jhh8:
converted to clientside, made it account for hidden weapons, level requirements etc. only issue that i can think of is that it takes into account the previous loadout, so for example if you have a minigun in your secondary slot and run this command, it will not be able to give you the minigun to your primary slot. despite this small issue it works good from my tests, i like it

PR #659, opened 2023-10-31
fix gas grenade particles disappearing

jhh8:
make gas grenade particles fully client side, just like firemine and env_fire particles


jhh8:
there's one small problem i realised, if a player joins the server while the gas has been up for lets say 20 seconds, then it will show the particles to that person for 30 seconds instead of the 10 seconds that are left, i assume? the solution could be to have a float netprop which tells at which time the grenade has detonated, then do StopEmission

BenLubar:
This will break if someone throws a gas grenade after a level has been running for ~12 days but I think that's fine.

PR #663, opened 2023-11-21
Updated challenge_asw_infection.nut

ModdedMarionette:


reactivedrop/content/infection/scripts/vscripts/challenge_asw_infection.nut
```diff
@@ -1013,6 +1013,7 @@ function UseLastStand(hMarine)
 	g_lastHuman[hMarine] <- hBubble;
 	hMarine.SetHealth(GetNewHealth(hMarine));
 	ClientPrint(null, 3, NameFeed(hMarine) + " activated Last Stand mode.");
+	ClientPrint(null, 3, "+60 seconds to the match timer.");
```
BenLubar, line 1016: We don't have to do this right now, but it'd be good to add these to `reactivedrop_english.txt` and reference the messages by ID. For the `NameFeed(hMarine)` one, you can do `ClientPrint(null, 3, "#some_string_name", NameFeed(hMarine))` and have a `%s1` in the translation string where the name should go.

PR #674, opened 2023-11-29
healamp gun: dont use ammo if marine cant be buffed, don't show buff …

jhh8:
…effect too


jhh8:
would be nice to hide the laser pointer when doing secondaryattack just like it hides when doing normal healing, i didnt go deep enough to figure out how to do it

PR #691, opened 2023-12-17
integrate 8 player list lobby ui addon

jhh8:


BenLubar (changes requested):
I'd rather use something like 2024 instead of "new" so we don't go down the route of Nintendo and have "new super mario bros" that is like 10 revisions ago.

PR #693, opened 2023-12-18
add hardcore ff to difficulty string in briefing

jhh8:
for consistency will need to add nb_hardcore_title string instead of using the other one

### DIFF
--- a/src/game/client/swarm/vgui/nb_mission_summary.cpp
+++ b/src/game/client/swarm/vgui/nb_mission_summary.cpp
@@ -123,16 +123,10 @@ void CNB_Mission_Summary::OnThink()
 		pDifficulty = L"";
 	}
 
-	if ( CAlienSwarm::IsOnslaught() )
-	{
-		wchar_t wszText[ 128 ];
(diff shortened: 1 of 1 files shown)


src/game/client/swarm/vgui/nb_mission_summary.cpp
```diff
@@ -123,16 +123,10 @@ void CNB_Mission_Summary::OnThink()
...
-		V_snwprintf( wszText, ARRAYSIZE( wszText ), L"%s %s", pDifficulty, g_pVGuiLocalize->FindSafe( "#nb_onslaught_title" ) );
-		m_pDifficultyLabel->SetText( wszText );
-	}
-	else
-	{
-		m_pDifficultyLabel->SetText( pDifficulty );
-	}
+	wchar_t wszText[ 128 ];
+	//                                                feels like its weird to do it like this
+	V_snwprintf( wszText, ARRAYSIZE( wszText ), CAlienSwarm::IsHardcoreFF() ? L"%s %s %s" : L"%s%s %s", CAlienSwarm::IsHardcoreFF() ? g_pVGuiLocalize->FindSafe("#nb_hardcore_title") : L"", pDifficulty, CAlienSwarm::IsOnslaught() ? g_pVGuiLocalize->FindSafe("#nb_onslaught_title") : L"");
```
BenLubar: I'd rather use an if statement here than two inline ternary expressions with the same condition.

PR #695, opened 2023-12-25
clientside grenade trail particles

jhh8:
fixes GL and vindicator firenade particles not appearing with packet loss, uses same strategy as rockets from hornet barrage/smartbomb

### DIFF
--- a/src/game/client/swarm/c_asw_grenade_vindicator.cpp
+++ b/src/game/client/swarm/c_asw_grenade_vindicator.cpp
@@ -1,0 +1,37 @@
+#include "cbase.h"
+#include "c_asw_grenade_vindicator.h"
+
+// memdbgon must be the last include file in a .cpp file!!!
+#include "tier0/memdbgon.h"
+
+IMPLEMENT_CLIENTCLASS_DT( C_ASW_Grenade_Vindicator, DT_ASW_Grenade_Vindicator, CASW_Grenade_Vindicator )
+	
+END_RECV_TABLE();
+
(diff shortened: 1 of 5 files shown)


jhh8:
next thing i'd like to do is have explosions also always show up. I was thinking to create explosions in UpdateOnRemove, however that makes them not accurate with higher packetloss and/or lower updaterate. It would be accurate to send the explosion origin together with the remove update, im not sure if simply having a vecExplosionOrigin netprop and setting it right before UTIL_Remove would acheive that, need to test

jhh8:
nevermind the accurate explosion particles idea above, people are saying they would rather not see explosion particles when theres packet loss just because their lack of presence is a good indication that there is packet loss. for future reference, the way i did it for grenade launcher grenades explosion was to have a m_vecDetonateOrigin netprop and set it to the origin of the grenade when it detonates, checking in client when that netprop becomes a non-default value, then simulate the particles. since the grenade gets removed right after the detonation, this vector wont be sent to the client, so i delayed the grenade removal by a short amount like 0.5 to 1.0 seconds at the end of the Detonate function by doing Disable(), then SetThink(SUB_Remove), SetNextThink(gpGlobals->curtime + 1.0f). Haven't gotten into the way of making the grenade invisible while still sending over that netprop, because with SetEFlags(EF_NODRAW) it doesnt want to send it anymore. An easier way would be to make ASWGrenadeExplosion messages reliable, however the explosion sound is a seperate message which i havent checked if can or should be made reliable. I'd rather not have those messages as reliable though

PR #709, opened 2024-01-13
dont reset leader/kick votes on instant restart, add a instant restar…

jhh8:
…t cooldown

meant to help against lobby leader abusing instant restart

move m_iLeaderVotes and m_iKickVotes from game resource to gamerules

limit cooldown cvar max to 10 in case of some sort of abusive addon trying to set it to a very high value. notify flag for same reason


BenLubar:
I think we can not have a limit for it and just add it to the list of non-challenge convars like hostname.

jhh8:
cant it still be changed with Convars.SetValue or SendToConsole or point_servercommand?

BenLubar:
probably, but *someone* had to install the addon for it to do that, and instant restarts at times when you'd normally want to restart aren't rate limited

PR #734, opened 2024-02-29
Translatable vgui panels

Blueberryy:
Didn't tested so i don't know if text supports non-english texts


BenLubar:
I don't think we should be translating debug screens.

Blueberryy:
> I don't think we should be translating debug screens.

They have issues with font for other languages?

Blueberryy:
I personally want to translate this so i hope it will not break anything while showing

PR #750, opened 2024-03-30
Infection Patch 4

ModdedMarionette:
Leaving this as draft until before the update, in case I have any other changes to make.


BenLubar:
gonna make the per-npc speed setting work if you want to switch to that

ModdedMarionette:
> gonna make the per-npc speed setting work if you want to switch to that

"m_flLaggedMovementValue" changes gravity as well. It's like it's slowing down time for the movement entirely.
Would you be making it so it changes the base speed per marine like how the speed skill bonuses do it?

ModdedMarionette:
@BenLubar 
Works no problem. Change has been pushed.

PR #853, opened 2025-02-16
constantly search for briefing camera

jhh8:


BenLubar:
Hmm, hopefully this won't affect perf too much.

PR #857, opened 2025-02-23
fix vindinade particles and sound disappearing with packet loss

jhh8:


jhh8:
gonna do for GL nades soon

jhh8:
ran into some complications with GL nades, maybe another time. vindi nades only for now

PR #858, opened 2025-02-23
Die() vscript function for inhabitable npcs

jhh8:
current way to kill something is a little more complicated than it should be


ywgATustcbbs:
Would be nice if it is

Die(attacker = null)

jhh8:
Thats same as

attacker = null
Die(null)

ywgATustcbbs:
> Thats same as
> 
> attacker = null Die(null)

yeah. But Die(hMarine.entindex()) is useful in some cases

BenLubar:
> Would be nice if it is
> 
> Die(attacker = null)

I don't think we can do optional parameters in vscript binds

ywgATustcbbs:
> > Would be nice if it is
> > Die(attacker = null)
> 
> I don't think we can do optional parameters in vscript binds

In cpp:
 void CASW_Inhabitable_NPC::Die(TYPE  attacker = null){...}

I think this will work. At least it allows to insta kill a marine and giving the attacker info

jhh8:
Use existing TakeDamage function for this

PR #861, opened 2025-03-24
cl_showspectators

jhh8:


src/game/client/vgui_fpspanel.cpp
```diff
@@ -517,6 +520,57 @@ void CFPSPanel::Paint()
...
+			int nSpecCount = 0;
+			for ( int id = 1; id < 31; id += 1 )
+			{
+				if ( !g_PR->IsConnected( id ) )
+					continue;
+
+				if ( g_PR->IsFakePlayer( id ) )
+					continue;
+
+				if ( !( pPlayer->m_iSpectatorIndexes.Get() & static_cast<int>( pow( 2, id ) ) ) )
```
BenLubar:
```suggestion
				if ( !( pPlayer->m_iSpectatorIndexes.Get() & ( 1 << id ) ) )
```

src/game/server/swarm/asw_player.cpp
```diff
@@ -2213,7 +2218,23 @@ void CASW_Player::SpectateNextMarine()
 
 void CASW_Player::SetSpectatingNPC( CASW_Inhabitable_NPC *pSpectating )
 {
+	int nOwnIndex = pow( 2, GetClientIndex() + 1 );
```
BenLubar: use `<<`, `|=`, and `&=~`

src/game/server/swarm/asw_player.h
```diff
@@ -91,6 +91,7 @@ class CASW_Player : public CBaseMultiplayerPlayer, public IASWPlayerAnimStateHel
 	CASW_Inhabitable_NPC *GetSpectatingNPC() const;
 	HSCRIPT ScriptGetSpectatingNPC() const;
 	CNetworkHandle( CASW_Inhabitable_NPC, m_hSpectating );
+	CNetworkVar( int, m_iSpectatorIndexes );
```
BenLubar: should be unsigned if we're allowing a 32nd player

jhh8: Ive looked through some examples how other classes define unsigned int and theyre still an int but in sendinfo they send a sprop_unsigned, didnt get it

BenLubar: You can replace `int` with `unsigned` and it should work fine.

jhh8: hm also it should work either way, 31st player will be -2147483648 instead of 2147483648 but bitwise operations will still work correctly (32nd player is the one becoming 1)

PR #872, opened 2025-03-30
Fixes: Maps: rd-tft2AbandonedMaintenance

Kitsune44:
[Fixes.](https://discord.com/channels/310381338538541056/1041816036439236650/1355896354697777356)

In the room with the energy beam:
- fixed an issue where the blockade of the beam crossing was triggered too quickly – it is now activated only after the biomass ignited by the beam has burned out.
- replaced the window passage with stairs
- added a warning info_overlay before the stairs leading to the beam
- added a hint(#rd_weapon_fact_damage_per_shot_beam_after) warning about the beam (appears only when the player is at a certain distance from the beam – when they step onto the stairs)
- changed the beam color to red (RGB: 240 30 30)
- changed the beam receiver's indicator lights to red (RGB: 240 30 30)

rd-tft2abandonedmaintenance
rd-tft2abandonedmaintenance_old

### DIFF
--- a/contentsrc/mapsrc/rd-tft2AbandonedMaintenance.vmf
+++ b/contentsrc/mapsrc/rd-tft2AbandonedMaintenance.vmf
@@ -2,7 +2,7 @@ versioninfo
 {
 	"editorversion" "400"
 	"editorbuild" "7397"
-	"mapversion" "1350"
+	"mapversion" "1376"
 	"formatversion" "100"
 	"prefab" "0"
 }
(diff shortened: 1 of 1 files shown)


WhenTanksFly:
Hmmm the light red glow may be less visible on yellow ground than cyan glow.
How much does it change depending on graphics settings, specifically Bloom that is High by default? (in-game: Esc -> Options -> Video -> Advanced settings -> Bloom; not sure where is it located in the settings menu accessible from the game's main menu)
glow mean good
High vs. None

Kitsune44:
Changing the color to red is a suggestion from Discord, rightly argued by the fact that red has strong connotations with danger, and on this map, many players walk into the blue laser beam and die. Following the discussion on Discord, I wanted to make the danger more clearly signaled without preventing death in the beam.
Btw, the platform is not yellow, it is grey, its colour comes from lighting - from the general lighting of the room which is in a brown-yellow tone and from lighting with a beam which is the same colour as the beam.
I will not comment on the glow/bloom values, I did not change them.

PR #877, opened 2025-04-02
Fixes: Maps: Disabled friendly fire of dropship sentries. 

Kitsune44:
Disabled friendly fire of dropship sentries on maps:
- ASI-Jac2-Deima
- ASI-Jac7-TimorStation
- rd-til7Factory

Discussed on Discord.


ywgATustcbbs:
o.o
That means i need to set ff on somehow in Add-on in the future. I need gunship to make mission more chaotic and dangerous

PR #880, opened 2025-04-04
Fixes: Maps: ASI-Jac7-TimorStation.

### LINKED ISSUE #878: Jacob's Rest - Timor Station: Preemptive collision at destination of small bridge by north hack
A noob's standard procedure is to ignore south eggs (maybe even the east huge egg pile as well) and proceed to the binary hack. The south eggs left alone often have their parasites walk to this mid-air emptiness instead of jumping towards marines right away.
Image
The bridge itself already seems to have proper marine/alien/projectile collision, so this hardened air doesn't seem to have any good purpose.

Kitsune44:
Fixes #878.


jhh8:
Are you disabling/killing the npc clip when bridge opens?

Kitsune44:
> Are you disabling/killing the npc clip when bridge opens?

Yes, it is not visible in the changes because I used the existing mechanism:
image

PR #882, opened 2025-04-05
fix flamer and shotgun ignoring asw_auto_reload 0 on secondary attack

jhh8:
also most weapons have this useless if statement which checks whether you have fired a weapon with 0 ammo (impossible) but i wont be the one to remove those in case theres some hilariously absurd scenario where that can happen somehow

### DIFF
--- a/src/game/shared/swarm/asw_weapon_flamer_shared.cpp
+++ b/src/game/shared/swarm/asw_weapon_flamer_shared.cpp
@@ -190,19 +190,22 @@ void CASW_Weapon_Flamer::ItemPostFrame( void )
 	BaseClass::ItemPostFrame();
 
 	bool bAttack1, bAttack2, bReload, bOldReload, bOldAttack1;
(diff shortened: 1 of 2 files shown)


jhh8:
Hm okay the "hilariously absurd acenario" could be cheaters with custom client dlls i guess

BenLubar:
If you're firing the extinguisher and you don't want it to reload automatically, why would you lose the last bit of ammo?

jhh8:
Do you suggest we fire 1 more exringuisher pellet at cost of 1 ammo instead of 2?

jhh8:
Or if you mean make it not fire at all, hm i guess

jhh8:
but now theres another issue with flamer which need to fix, starting to hold primary attack when on 1 ammo makes it fire once, then server stops firing but client still thinks its firing and emits sound. same for secondary attack but also particles

jhh8:
this issue is present on current build also

PR #887, opened 2025-04-10
fix point_spotlight looking weird when being moved

jhh8:
theres so many more broken things about this entity... blame 1950s code


jhh8:
https://github.com/user-attachments/assets/39c82443-f189-47ae-a85d-eb3952e37edd

before and after

PR #899, opened 2025-05-25
restrict cursor to 16:9 area to prevent ultra-wide fov cheat

Victor-Diego:


ywgATustcbbs:
There are many screen ratios such as 16:10 4:3 1:1 9:16 etc. You might want to consider these cases.
Technically if one rotates the screen 90 degrees to 9:16, he will have extra top view.

Victor-Diego:
> There are many screen ratios such as 16:10 4:3 1:1 9:16 etc. You might want to consider these cases. Technically if one rotates the screen 90 degrees to 9:16, he will have extra top view.

Alien Swarm does not append extra field of view along Y axis, thanks for the suggestion anyway

PR #900, opened 2025-05-26
restrict cursor to 16:9 area to prevent ultra-wide fov cheat

Victor-Diego:
also added two new convars:
rd_draw_restricted_borders (def="1") -> Display the restricted cursor area when using ultra-wide resolution
rd_draw_restricted_borders_color (def="128 128 128 128") -> Color of the restricted cursor area borders
image

### DIFF
--- a/src/game/client/swarm/asw_in_mouse.cpp
+++ b/src/game/client/swarm/asw_in_mouse.cpp
@@ -15,6 +15,9 @@
 // memdbgon must be the last include file in a .cpp file!!!
 #include "tier0/memdbgon.h"
 
+bool g_ultra_wide_screen = false;
+Rect_t g_clamp_area = {};
+
 ConVar glow_outline_color_active( "glow_outline_color_active", "153 153 204", FCVAR_NONE );
 ConVar glow_outline_color_inactive( "glow_outline_color_inactive", "77 77 77", FCVAR_NONE );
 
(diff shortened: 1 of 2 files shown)


Victor-Diego:
For testers:
Alien Swarm Reactive Drop\reactivedrop\cfg\video.txt <- Modify the file to customize game resolution
Borders should be displayed only when using resolution wider than 16:9

src/game/client/swarm/vgui/asw_hud_master.cpp
```diff
@@ -480,6 +486,17 @@ void CASW_Hud_Master::Paint( void )
 
 	if ( m_pLocalMarineResource )
 	{
+		// draw restricted borders for ultra-wide screen
+		if (rd_draw_restricted_borders.GetBool() && g_ultra_wide_screen && !pPlayer->IsObserver()) {
```
BenLubar: spectating is determined in AS:RD by whether the player has a non-null SpectatingNPC

src/game/client/swarm/asw_in_mouse.cpp
```diff
@@ -85,6 +88,36 @@ void CASWInput::ApplyMouse( int nSlot, QAngle& viewangles, CUserCmd *cmd, float
 	int current_posx, current_posy;
 	GetMousePos(current_posx, current_posy);
 
+	// restrict cursor to 16:9 area to prevent ultra-wide fov cheat
+	g_ultra_wide_screen = false;
+	C_BasePlayer* player = C_BasePlayer::GetLocalPlayer();
+
+	if (player && !player->IsObserver()) {
```
BenLubar: spectating is determined in AS:RD by whether the player has a non-null SpectatingNPC

BenLubar:
> For testers: Alien Swarm Reactive Drop\reactivedrop\cfg\video.txt <- Modify the file to customize game resolution Borders should be displayed only when using resolution wider than 16:9

You can also start AS:RD with `-w 1920 -h 540` on the command line for 32:9

src/game/client/swarm/asw_in_mouse.cpp
Victor-Diego: got it

PR #906, opened 2025-06-15
Fixes: VScript: CASW_Mission_Chooser_VScript.

Kitsune44:
Fix map transition crashes by using CreateTable instead of CreateArray.


Kitsune44:
@BenLubar

BenLubar:
Does ArrayPush work on a table? Arrays are different from tables in Squirrel.

PR #909, opened 2025-06-17
Fix ultra wide screen calculation mistakes introduced by refactoring. Bring minimap to front

ywgATustcbbs:


ywgATustcbbs:
A silly mistake.

PR #910, opened 2025-06-24
UI/UX improvements

ywgATustcbbs:
- [x]  Add a client side vscript function for CRD_Vgui_VScript to force syncing with server
- [x] Revise wiki page to better describing behaviors of GetInt() / SetInt() functions of vgui entity
- [x] Rename cl_add_index_to_name to rd_add_index_to_name
- [x] Improve user name displaying. For both client and server side
- [x] Show secret marks on Traitor marines for Traitor team
- [x] Add new Traitor icons.
- [x] Minor fixes


ywgATustcbbs:
Will edit wiki after being merged

I don't know how to side load textures for debugging. Therefore those textures might be broken.

jhh8:
could you move the 3 functions related to UTF strings into a utils header and include that header (if its not already included) instead of having 3 copies of the functions?

ywgATustcbbs:
> could you move the 3 functions related to UTF strings into a utils header and include that header (if its not already included) instead of having 3 copies of the functions?

tried to put them into util header. They can't pass build process. Very strange. Don't know why.

I found They best fit into strtools.h. However if i put them there,

ywgATustcbbs:
Ok. Now put utf8 functions into util_shared.h

BenLubar:
I'm going to delete the png files after merging this so they aren't included as loose files in the game depot.

PR #918, opened 2025-07-19
fixes a 2-5 second freeze interval on the main menu

mithrand0:
this patch fixes a possible 2-5 seconds freeze on the main menu.

the freeze comes from scanning *a lot* of lobbies, not only as:rd lobbies, but *all* internet lobbies. This comes forth from SteamMatchmaking, which does not apply an AppID filter by default.

This patch ensures only lobbies with gametype "reactivedrop" are added to the list. This limits the amount of udp calls further down the sequence, since only as:rd lobbies are scanned for details, instead of all lobbies.

not exactly sure why the lobbies are all scanned with udp, but at least this patch ensures only as:rd lobbies are scanned.


mithrand0:
Ben says it's Valve's fault, so let's close it for now.

PR #924, opened 2025-07-27
Add 'lock-mode' for Devastator

geotavros:
This commit adds 2 new ConVars

- rd_devastator_lockmode_enabled (default is 1) Enables lock-mode secondary attack. Marine is locked in place but weapon fire rate and penetration gets higher.
- rd_devastator_lockmode_firerate (default is 1.3) Scale factor of the fire rate in lock-mode

When lock-mode is enabled using the ConVar above then the following functionality is enabled for Devastator:
- Pressing Secondary Fire button locks marine in place
- Devastator's fire rate is increased by rd_devastator_lockmode_firerate factor.
- Devastator's pellets will penetrate 1 alien with a chance of 100%
- Pressing Secondary Fire button again will disable lock-mode
- Things that disable lock-mode
  - Roll
  - Weapon swap or weapon drop
  - Melee attack
  - Activating jump jet or blink
  - If damaged

### DIFF
--- a/reactivedrop/scripts/asw_weapon_devastator.txt
+++ b/reactivedrop/scripts/asw_weapon_devastator.txt
@@ -24,6 +24,7 @@ WeaponData
 	{
 		"empty"					"ASW_Weapon.Empty"
 		"special3"				"ASW_Weapon.Reload3"				// weapon switch
+		"burst"					"ASW_Weapon.Reload2"
 
 		"single_shot"			"ASW_Weapon_Devastator.SingleFP"		// sound for locally controlled marine
 		"single_shot_npc"		"ASW_Weapon_Devastator.Single"		// sound for other marines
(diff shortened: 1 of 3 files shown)


jhh8:
Why not have it by default

geotavros:
The idea was to playtest it in challenges first to gather players' feedback, fix bugs and improve, for example in Weapons Balancing challenge.

But I'm not against making it enabled by default too.

geotavros:
Players' feedback so far:
- Add a sound effect when lock-mode gets disabled
- Add a particle effect when lock-mode is enabled

GmaOrange:
Would be nice to have some kind of visuals (particles\sprites etc) because visual way faster than sound to react from.
Rest seems ok. It doesnt make this gun super usefull still but doesnt ruin either so why not.

GmaOrange (approved):

ax-now:
Would be cool to have some visual inducator on crosshair when u are in lock-mode. 
A visual on the weapon is cool, but generally you look at your sights and what you're shooting at. The marksman has its scope, the medgun has its healing icon. Since the weapon interacts with your movement ability, additional information on the sights would be welcome, what do you think?

PR #931, opened 2025-08-05
fix a bug where the challenge is reset after asw_mission_restart

mithrand0:


geotavros:
rd_challenge ConVar has a callback UpdateMatchmakingTagsCallback(), which runs a lot of code inside of it. It needs to be reviewed to understand if it's safe to change rd_challenge like this and in what place. Sorry, but I don't have time right now to review it.

mithrand0:
> rd_challenge ConVar has a callback UpdateMatchmakingTagsCallback(), which runs a lot of code inside of it. It needs to be reviewed to understand if it's safe to change rd_challenge like this and in what place. Sorry, but I don't have time right now to review it.

Thanks for pointing out UpdateMatchmakingTagsCallback(), I'll dive into it.

mithrand0:
I reverted pull request #926, and added a convar, allowing me to check for this specific state.
In my opinion, this should eventually be fixed correctly, but this way it allows me to continue my other work.

TLDR: the functionality is reverted to what is was before #926

PR #932, opened 2025-08-06
Make Medical SMG more accurate when crouching

geotavros:
Similarly to how Minigun works, Medical SMG bullet spread is decreased when marine is crouching.
Currently set to 6 degrees like Minigun.

This makes weapon slightly more useful in killing single aliens which are far from marine.


GmaOrange (approved):

mithrand0 (approved):

BenLubar:
The Swarmopedia needs to be updated with a weapon fact for this.

PR #936, opened 2025-08-17
add missing challenge.setvalue

mithrand0:
During an earlier revert, this line went missing. 

Copied the file from an earlier version, and the file works now as before.


mithrand0:
Looks like there were changes in-between, need to take a better look..

Issue #935, opened 2025-08-08
Add an icon for "Thanks" emote


geotavros:
Right now when executing "Thanks" emote this results in marine speaking "Thank you". We could add an icon similar to "Go" and "Stop" emotes.

geotavros:
Here's how it's going to look like for now: 

<img width="1544" height="869" alt="Image" src="https://github.com/user-attachments/assets/b936d779-9f0c-4746-9ab5-2fbdb41db776" />

GmaOrange:
i miss native trollface

PR #945, opened 2025-08-21
Add "Thanks" emote

geotavros:
When choosing "Thanks" in the radial menu marine will now show a blue filled circle with "Thanks" text inside of it as well as chatter "Thanks".

See #935 for more details.
"cl_emote 14" command can be used to bind this emote to some key.


BenLubar:
This seems like a great use case for the LanguagePreference proxy.

PR #953, opened 2025-09-01
added seal icon to radial menu "request seal" item

hpdodger:
image


geotavros:
Using the same texture as the welding icon is a little bit confusing. I think a new texture for the emote would be a better choice.

src/game/server/swarm/asw_marine.cpp
```diff
@@ -4496,6 +4496,9 @@ void CASW_Marine::DoEmote( int iEmote )
 		case 14:
 			GetMarineSpeech()->Chatter( CHATTER_THANKS );
 			break;
+		case 15:
+			// just request seal
```
geotavros: CHATTER_REQUEST_SEAL_DOOR can be used here. Otherwise it will just be silence?

hpdodger:
> Using the same texture as the welding icon is a little bit confusing. I think a new texture for the emote would be a better choice.

asked AstroCatNip for help to create welder emote

src/game/server/swarm/asw_marine.cpp
hpdodger: yes, thank you

hpdodger:
image

new icon ai generated

BenLubar:
We can't use AI-generated content.

hpdodger:
image

src/game/client/swarm/vgui/asw_hud_emotes.cpp
```diff
@@ -72,7 +72,7 @@ class CASWHudEmotes : public CASW_HudElement, public vgui::Panel
 	CPanelAnimationVarAliasType( int, m_nSentryUpTexture, "SentryUpTexture", "vgui/swarm/ClassIcons/SentryBuild", "textureid" );
 	CPanelAnimationVarAliasType( int, m_nSentryDnTexture, "SentryDnTexture", "vgui/swarm/ClassIcons/SentryDismantle", "textureid" );
 	CPanelAnimationVarAliasType( int, m_nHackTexture, "HackTexture", "vgui/swarm/ClassIcons/HackIcon", "textureid" );
-	CPanelAnimationVarAliasType( int, m_nWeldTexture, "WeldTexture", "vgui/swarm/ClassIcons/WeldIcon", "textureid" );
+	CPanelAnimationVarAliasType(int, m_nWeldTexture, "WeldTexture", "vgui/swarm/Emotes/EmoteWelder", "textureid");
```
geotavros: I assume this change also changes that icon for when the actual welding happens. To fix this issue you need to create a new texture variable instead of using m_nWeldTexture

hpdodger:
thank you for patience, struggling with unfamiliar stack

geotavros:
Looks good from quick code review side. I didn't test it in-game. If @hpdodger tested it and it works fine then I think it can be merged.


`SendPropInt		( SENDINFO( m_iEmote ), 16, SPROP_UNSIGNED ),`

I assume with the current approach we're limited to 16 emotes max, so this will be the last one available.

BenLubar:
> I assume with the current approach we're limited to 16 emotes max, so this will be the last one available.

The number can be increased up to 32, but until we're using all the bits, there's no point in doing so.

Plus, today's hotfix freed up all the traitor mode emote slots.

geotavros:
@BenLubar Do you plan to merge this one for next update?

PR #954, opened 2025-09-01
Disable black bars by default on coop. Enbales by default on deathmatch.

ywgATustcbbs:
This is a walkaround before final resoluation fix.

Simply set ConVar related to ultra wide screen black bar to 0. 

The code is not removed Since Traitors challenge wants to block extra fov.


WhenTanksFly:
From #900 I thought there'd be no black bars in the first place, just restricted cursor area. For clarification: does this change revert to exactly that behavior for normal coop games, or does it undo the whole protection?

PR #957, opened 2025-09-04
Fix Reduction - Fallen City: Fake, non-igniting fire in egg zone near end

### LINKED ISSUE #844: Reduction - Fallen City: Fake, non-igniting fire in egg zone near end
https://github.com/user-attachments/assets/8ab69b08-cd9f-4213-90e1-c7115d0378b1

Once it's fixed, time of fire appearance there may have to be adjusted to not auto-kill the parasite spawn. Or just remove the fire, idk.

geotavros:
The fire actually ignites but it takes longer time, non-instant.

Setting flags to Infinite Duration makes fire ignite instantly.

Fixes #844


Kitsune44:
Fixes #844.

PR #960, opened 2025-09-05
update traitors challenge - fix client crash

ywgATustcbbs:
The bug seems to be the pointer in Script_PaintPolygon that used by CUtlMemory<vgui::Vertex_t> later. It seems that CUtlMemory is the key problem -- it accesses the pointer's values via AssignTo() function. Therefore, wrote a new hud function to avoid passing pointers.


BenLubar (changes requested):
I'm pretty sure `PaintTexturedRectangle` would work here, no need to add a new function.

PR #963, opened 2025-09-09
Do not draw aliens outside 16:9 area of the screen.

ywgATustcbbs:
This pr only blocks extra view. Egg is not an alien type. No idea if it should be fixed. Therefore, leave it as is.
Needs another pr to block jump jets, blink, nade etc.

Tested on my client. Works fine.

### DIFF
--- a/src/game/client/clientleafsystem.cpp
+++ b/src/game/client/clientleafsystem.cpp
@@ -302,6 +302,8 @@ private:
 
 	void CalcRenderableWorldSpaceAABB_Bloated( const RenderableInfo_t &info, Vector &absMin, Vector &absMax );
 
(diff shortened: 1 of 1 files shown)


ywgATustcbbs:
THis is kinda prototype pr. Due to the current state of the discussions, No control logic is added.

Control logic could be added easily after it comes to a final agreement.

In case the returned value of ShouldRenderableBeIgnored() Is modified by CE or something, it should be protected.

mithrand0:
Since the proposed changes aim to limit cheating, they will also require substantial additional work in one of our closed-source repositories. For that reason, this PR (and all black-bar, visibility or aim restrictions PR's in general) should not be merged until both components are implemented together. 

I propose to convert this PR to a draft, to prevent an accidental merge.

PR #966, opened 2025-09-21
Add bonus mission Reversed Syntek Residential

geotavros:


BenLubar:
we still need deima added to build_maps.py and the overview folder

PR #986, opened 2025-10-20
Improved auto close function

connman116:
Changes to  
asw_door.h
asw_door.cpp
BasePropDoor.h

moved autoclose from props.cpp to asw_door.cpp and improved it in the process to detect marines near doors and reschedule autoclosing if it attempts to autoclose when a marine is near a door

removed doorclose from asw_door.h when theres no door close function in asw_door.cpp


ModdedMarionette:
This is a band-aid workaround fix that creates an inconsistency between the doors' open logic and close logic.
Please hold off on pushing this change, as we are far from the point of having to resort to workarounds with guaranteed side effects.

PR #987, opened 2025-10-20
make doors not try to close when something is in the door area

### LINKED ISSUE #12: door closing
door forcibly tries to close after a few seconds of being open even if something is standing in the door area the whole time

jhh8:
fixes #12

### DIFF
--- a/src/game/server/swarm/asw_door.cpp
+++ b/src/game/server/swarm/asw_door.cpp
@@ -605,6 +605,13 @@ bool CASW_Door::CheckDoorClear()
 		return false;
 	}
 
+	// dont try to close when an entity is standing inside any trigger_asw_door_area's that would make the door open
+	FOR_EACH_VEC( m_AttachedDoorAreas, i )
+	{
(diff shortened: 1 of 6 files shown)


Kitsune44 (approved):
- [x] Testing / Verification

2arnold888:
Hello, would be nice if we could add some loop to door area. Like add a timer to a door area to keep it open when a player shows up and leaves to keep it open and if someone else enters the area to reset time to keep the door open. From g play perspective the player interacts with a door it opens than he backs up to cover and the door closes when he tries to pass again, in other cases if a squad is moving one after another and there is bigger gap between players door closes right in front and separates member of a squad and delays him. In asbi RNG those cases means life n death.

-Best regards

Kitsune44:
Hello @2arnold888,

The doors already have such a mechanism. The default time after which they close is 5 seconds, and it can be set individually for each door by the map creator.
The extension this PR introduces is complementary to that system - the timer still works, but the doors won’t close if someone is standing in their trigger area(imagine the detection zone of a door’s photo eye in the game world).

-Best regards

PR #1006, opened 2025-11-18
replace acute symbol with quote symbol when parsing a runscript string

jhh8:
so that you can write the acute symbol in hammer runscriptcode parameter field which gets turned into a quote symbol where otherwise writing a quote symbol directly breaks the vmf parsing in hammer


BenLubar:
we TF2 now

PR #1007, opened 2025-11-19
Update asw_marine.cpp make hornets and smartbomb nolonger friendlyfire

connman116:
As polled on discord make hornets and smart bomb nolonger friendly fire

### DIFF
--- a/src/game/server/swarm/asw_marine.cpp
+++ b/src/game/server/swarm/asw_marine.cpp
@@ -1419,7 +1419,12 @@ int CASW_Marine::OnTakeDamage_Alive( const CTakeDamageInfo &info )
 				Msg( "  but all ignored, since it's from a team mate\n" );
 			return 0;
 		}
-		else
+		//make hornet and smart bomb no longer friendlyfire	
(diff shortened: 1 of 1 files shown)


src/game/server/swarm/asw_marine.cpp
```diff
@@ -1419,7 +1419,12 @@ int CASW_Marine::OnTakeDamage_Alive( const CTakeDamageInfo &info )
 				Msg( "  but all ignored, since it's from a team mate\n" );
 			return 0;
 		}
-		else
+		//make hornet and smart bomb no longer friendlyfire	
+		else if (!ASWDeathmatchMode() && pWeapon && (pWeapon->Classify() == CLASS_ASW_HORNET_BARRAGE || pWeapon->Classify() == CLASS_ASW_SMART_BOMB))
+			{
+    			return 0;
+			}			
+else
```
BenLubar:
```suggestion
		else
```

it should be over there with its friends

jhh8:
weapon is deleted when the last rockets are fired, so weapon becomes null and those rockets will deal damage, its more reliable to check for inflictor to be the rocket

PR #1009, opened 2025-12-10
cvarlist_rd concommand

jhh8:
works on server, preferably need someone to test on client (client compilation doesnt work for me since a year ago, lazy to solve)

### DIFF
--- a/src/game/client/swarm/rd_convar_hacks.cpp
+++ b/src/game/client/swarm/rd_convar_hacks.cpp
@@ -190,3 +190,89 @@ public:
 		pConVar->SetValue( szNewDefault );
 	}
 } s_RD_Convar_Hacks;
+
+
+
+static bool ConCommandBaseSortFunc( const ConCommandBase* const &hLeftCMD, const ConCommandBase* const &hRightCMD )
+{ 
+	const char* szLeftCMDName = hLeftCMD->GetName();
+	const char* szRightCMDName = hRightCMD->GetName();
+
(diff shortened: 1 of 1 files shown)


src/game/client/swarm/rd_convar_hacks.cpp
```diff
@@ -190,3 +190,89 @@ static class CRD_Convar_Hacks final : public CAutoGameSystem
...
+
+				if ( pCMD->IsFlagSet( 1<<c ) && sizeof( s_szFlagDesc[c] ) != sizeof( "" ) )
+				{
+					Q_snprintf( szFlag, sizeof( szFlag ), ", %s", s_szFlagDesc[c] );
+					Q_strncat( szFullFlags, szFlag, sizeof( szFullFlags ), COPY_ALL_CHARACTERS );
+				}
+			}
+			
+			char tempbuff[512]{};
+			ConMsg( "%-40s : %-8s : %-16s : %s\n", pCMD->GetName(), dynamic_cast< const ConVar* >( pCMD )->GetString(), szFullFlags, pCMD->GetHelpText() );
```
BenLubar: shouldn't this print default values rather than current values if it's for documentation?

also if you're not checking whether the result is null and it's not some crazy multiple inheritance situation, there's no point in using `dynamic_cast` - use `static_cast` instead.

Kitsune44: I suggest displaying the current and default values ​​(`pCMD->GetDefault()`).

PR #1029, opened 2026-02-24
Fix steam activation requestexperience crash upon new player join on previously active server

tsiakoulias:
This old bug still plagued my classic Alien Swarm server, so i ported the fix on ASRD if you guys are interested.

(Brief bug description: Dedicated server launches, players connect, everything is smooth, players disconnect, server idles on lobby, a new player joins, server crashes upon player stats fetching).

You need to test it first, but I believe it will work without any issues.

Cheers and have fun!


tsiakoulias:
Replaced by a cleaner branch

PR #1049, opened 2026-05-10
Fix marines unexpectedly aiming at targets above the camera

### LINKED ISSUE #444: Unexpected vertical aiming
Imported from <https://steamcommunity.com/app/563560/discussions/2/6670425060425619954/>:

> **\-Ground aiming and autoaim:**
> It often happens that your character is aiming for the ground or it is aiming in air for no reason (i mean there must be a reason but it shouldn't be the case).
> For the aim in the air it is often caused by the auto aim which is much too capricious maybe it is also the case for the ground I don't know

We have two aiming modes: the new first/third person one that tries to rotate your gun to face the exact point you see in the middle of your screen, and the Alien Swarm top-down one that tries to aim flat but will aim up or down if your cursor is over an enemy. I'm hoping they're not mixing, but in either case, a demo would help me track this one down.

anf3is:
- Closes #444

This PR fixes an issue where marine aim could suddenly launch upward, even when there appeared to be nothing under the cursor.

The `rd_limit_aim_top_offset` convar has been introduced to allow customization of the aiming limit behavior.

## Description

The targeting logic now filters potential aim targets by comparing their Z coordinate against the camera's Z position. An additional offset lowers the threshold slightly to avoid cases where:
- a target's hitbox clips through the camera
- a target occupies most of the screen space

The change does not affect aiming at the ground and first person challenge and should not interfere with targeting enemies that are normally visible to the player.

## Offset examples

- **-50** - removes most noticeable aim glitches while still allowing targets positioned high above the player (for example, enemies on the cargo elevator ceiling)
- **-70** - ignores drones that are extremely close to the camera and fill roughly half the screen
- **-2000** - effectively disables aim target selection
- **99999** - effectively disables this fix

## Good maps for testing

- `as_city17_05` Citadel (last map of City 17) on the elevator
- `rd-bonus14_cargrev` Reversed Cargo Elevator
- `dc1-omega_city` Omega City (first map of Dead City) first trigger

---

P.S. A better fix would probably involve results of `UTIL_TraceLine`.

### DIFF
--- a/src/game/client/swarm/asw_input.cpp
+++ b/src/game/client/swarm/asw_input.cpp
@@ -55,6 +55,8 @@ ConVar joy_disable_movement_in_ui( "joy_disable_movement_in_ui", "1", FCVAR_NONE
 ConVar rd_input_controller_mode_mouse( "rd_input_controller_mode_mouse", "-1", FCVAR_NONE );
 ConVar rd_input_controller_mode_keyboard( "rd_input_controller_mode_keyboard", "-1", FCVAR_NONE );
 
+ConVar rd_aim_ignore_above_camera_z( "rd_aim_ignore_above_camera_z", "-60", FCVAR_NONE, "Targets above the camera by this Z distance are ignored in top-down mode" );
+
 extern kbutton_t in_attack;
 extern kbutton_t in_walk;
 extern int g_asw_iPlayerListOpen;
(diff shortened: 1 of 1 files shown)


anf3is:
It would probably be more intuitive if the offset was not negated. The convar description should then be reworded accordingly.
I'm also considering whether the default value should be higher.

PR #1052, opened 2026-05-15
Increase minigun low-ammo click sound threshold

### LINKED ISSUE #278: Minigun's low ammo sound threshold not updated
10ish ammo left is now too late to give any warnings at, it's gone in less than half a second. Let's 25-50?

anf3is:
- From default 6 to 8 (12 to 16) - increase by 33%. 25% would be better, but we work with int.
- Close #278


WhenTanksFly:
I've just re-checked the issue at hand, turns out I was wrong, please see the updated issue discussion thread.

Issue #590, opened 2023-05-14
Medical Amplifier Gun can be reloaded by Bastille, nulling all meds


WhenTanksFly:
It even correctly drops the clip on the ground.

BenLubar:
pushed to beta

PR #1063, opened 2026-05-19
Prevent reload of heal and amp ammo of various healguns

anf3is:
Fix issue where healguns would reload (to 0) when ammo count exceeded defined clip ammo count.

Should prevent any future regressions similar to #590, #1062.

### DIFF
--- a/src/game/shared/swarm/asw_weapon_healamp_gun_shared.h
+++ b/src/game/shared/swarm/asw_weapon_healamp_gun_shared.h
@@ -24,7 +24,7 @@ public:
 
 	virtual Class_T		Classify( void ) { return (Class_T)CLASS_ASW_HEALAMP_GUN; }
 	virtual bool		SecondaryAttackUsesPrimaryAmmo() { return false; }
-	virtual bool		UsesClipsForAmmo2() { return false; }
+	virtual bool		UsesClipsForAmmo2() const override { return false; }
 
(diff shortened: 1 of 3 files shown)


src/game/shared/swarm/asw_weapon_healamp_gun_shared.h
```diff
@@ -24,7 +24,7 @@ class CASW_Weapon_HealAmp_Gun : public CASW_Weapon_Heal_Gun
 
 	virtual Class_T		Classify( void ) { return (Class_T)CLASS_ASW_HEALAMP_GUN; }
 	virtual bool		SecondaryAttackUsesPrimaryAmmo() { return false; }
-	virtual bool		UsesClipsForAmmo2() { return false; }
+	virtual bool		UsesClipsForAmmo2() const override { return false; }
```
BenLubar, line 27: should this also have `UsesClipsForAmmo1`?

anf3is: `UsesClipsForAmmo1` is inherited from heal gun.
But I can add this to healamp gun too for clarity.

BenLubar: Ah, I forgot about the crazy weapon class tree.

PR #1064, opened 2026-05-19
Match heal ammo clip size to Faith's max ammo

anf3is:
Decrease from 100 to 90. Should affect displayed max count and ammo count of a default map spawned healgun.


BenLubar (changes requested):
Mapper defaults are in the fgd file.

(This will affect entities spawned by scripts, but the ones placed in hammer have the ammo as part of their definition.)

PR #1066, opened 2026-05-19
fix door health tooltip staying on screen after util_remove'd door

jhh8:
https://github.com/user-attachments/assets/649eff7c-18a5-4315-9e13-64efda43936b


BenLubar:
Cool, even more Warframe bugs that we're somehow managing to fix before DE can.

PR #1067, opened 2026-05-19
Prevent reloading when current ammo exceeds clip size

anf3is:
`CBaseCombatWeapon::DefaultReload` is not changed because all weapons and extras inherit from `CASW_Weapon` anyway.

Basic testing shows it works as intended, but it may need more testing.

### DIFF
--- a/src/game/shared/swarm/asw_weapon_shared.cpp
+++ b/src/game/shared/swarm/asw_weapon_shared.cpp
@@ -902,7 +902,7 @@ bool CASW_Weapon::ASWReload( int iClipSize1, int iClipSize2, int iActivity )
 	{
 		// need to reload primary clip?
 		int primary = MIN( iClipSize1 - m_iClip1, pMarine->GetAmmoCount( m_iPrimaryAmmoType ) );
-		if ( primary != 0 )
+		if ( primary > 0 )
 		{
 			bReload = true;
 		}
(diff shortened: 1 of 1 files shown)


src/game/shared/swarm/asw_weapon_shared.cpp
```diff
@@ -1200,7 +1200,8 @@ void CASW_Weapon::FinishReload( void )
 	if (pOwner)
 	{
 		// If I use primary clips, reload primary
-		if ( UsesClipsForAmmo1() )
+		if ( UsesClipsForAmmo1() &&
+				MIN( GetMaxClip1() - m_iClip1, pOwner->GetAmmoCount(m_iPrimaryAmmoType) ) > 0 )
```
BenLubar: This seems like a weird way to phrase this condition.

From what I can tell, you're checking both "do I have less than a full magazine currently loaded" and "is there reserve ammo available" and those should really be two separate expressions.

```suggestion
		if ( UsesClipsForAmmo1() && m_iClip1 < GetMaxClip1() && pOwner->GetAmmoCount( m_iPrimaryAmmoType ) > 0 )
```

anf3is: Ok, will do both for ammo1 and ammo2.

PR #1079, opened 2026-06-07
Add convar for spectate order

anf3is:
This PR aims to support custom spectating order (for example: Jaeger → Wildcat → Vegas → [other/fallback]).
The configured order should be respected when cycling between marines while preserving the current behavior that all alive marines remain valid spectate targets.

## WIP Info

- `SpectateNextMarine` is server-side. Because of that we need to send client's spectate order to server.
- `newmapsettings.cfg` is too early for `rd_set_spectate_order` :sob: 
- `rd_set_spectate_order` operates on `CASW_Player`. Do we need to call `rd_set_spectate_order` after client is fully connected and `CASW_Player` for this client is created? 

## Tasks

- [x] Add client-side `rd_spectate_order` for storing client's preferences
- [x] Find a place to automatically call `rd_set_spectate_order` if `rd_spectate_order` is set
- [x] Mark `rd_set_spectate_order` as hidden
- [x] Test in multiplayer


mithrand0:
We tested this in-game a while ago, it seemed to work back then.

anf3is:
I forgot about simple algorithms. You can find next (bigger) integer without sorting :facepalm:
Now need to learn how to send data from client to server in more resourceful way.

BenLubar:
It looks like you're doing the job that `FCVAR_USERINFO` can do for you.

PR #1081, opened 2026-06-11
add marineprofile keyvalue to info_player_start

jhh8:
marineprofile(choices) : "Marine" : -1 =
	[
		-1 : "Any"
		0 : "Sarge"
		1 : "Wildcat"
		2 : "Faith"
		3 : "Crash"
		4 : "Jaeger"
		5 : "Wolfe"
		6 : "Bastille"
		7 : "Vegas"
	]

### DIFF
--- a/src/game/server/swarm/asw_tutorial_spawning.cpp
+++ b/src/game/server/swarm/asw_tutorial_spawning.cpp
@@ -18,12 +18,12 @@ CASW_TutorialStartPoint *CASW_TutorialStartPoint::GetTutorialStartPoint( int iMa
 	if ( iMarineSlot < 0 || iMarineSlot >= 8 )
 		return NULL;
 
-	CASW_TutorialStartPoint *pStartEntity = assert_cast<CASW_TutorialStartPoint *>( gEntList.FindEntityByClassname( NULL, "info_tutorial_start" ) );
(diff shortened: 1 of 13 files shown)


jhh8:
as draft for now until i have time to test it a bit more

BenLubar:
Would probably rename "Default" to "Any"

src/game/server/swarm/asw_tutorial_spawning.cpp
```diff
@@ -18,12 +18,12 @@ CASW_TutorialStartPoint *CASW_TutorialStartPoint::GetTutorialStartPoint( int iMa
 	if ( iMarineSlot < 0 || iMarineSlot >= 8 )
 		return NULL;
 
-	CASW_TutorialStartPoint *pStartEntity = assert_cast<CASW_TutorialStartPoint *>( gEntList.FindEntityByClassname( NULL, "info_tutorial_start" ) );
+	CASW_TutorialStartPoint *pStartEntity = dynamic_cast<CASW_TutorialStartPoint *>( gEntList.FindEntityByClassname( NULL, "info_tutorial_start" ) );
```
BenLubar: `assert_cast` is cheaper at runtime than `dynamic_cast`, and it works as long as you can be sure that the value is either null or of the correct type, which should be the case here.

jhh8: it didnt want to compile

PR #1086, opened 2026-08-10
increase MAX_MAP_PLANES

jhh8:


jhh8:
need to build vrad_dll.dll for this to take effect

BenLubar:
Wouldn't this require rebuilding the engine? I don't think anything in client.dll or server.dll uses this directly.

jhh8:
its to not get numplanes + fakeplanes > MAX_MAP_PLANES error when compiling a map

jhh8:
hammer++ probably swapped to dynamic arrays to not have limits on planes/lights/brushes etc

jhh8:
To add more to this, theres also a brushsides limit which is the same 65536, launching a map with over that many brushsides (if you manage to compile it) crashes the game with "Map with too many planes". This num_max_planes is some seperate thing, havent looked into it in detail but feels like its a vrad only thing that has no references in the engine which counts things that dont get into compiled bsp like skip textured brushes or maybe it even counts displacement triangles. Because why it hits 65536 so much sooner than brushsides i dont know. I see it as just another random harmful source engine limit that has no real effect and limits brushsides even earlier than they should be

PR #1088, opened 2026-08-12
compatibilty for steam interface v9

mithrand0:
> Note: I have a **modified** engine.dll locally, that needs this. Without a patched engine, it won't work.

This PR is a draft, a proof of concept. It fully works, but is probably not production quality. It probably is good enough for a alpha or beta test.

A lot still needs to be tested, but it seems promising. Findings below are with the latest Steam SDK 1.65:

### Client
+ Client starts
+ Confirmed with latest `steam_api.dll`
+ Main screen seems to work
+ Workshop items from Steam do load
+ Can upload workshop submissions
+ Client can find and join lobbies
+ A map can be played and restarted

### Server
+ Server starts
+ Confirmed with latest `steam_api.dll`
+ Steam API is inited from game code instead of engine
+ Gameserver is inited from game code instead of engine
+ Heartbeat to master servers is sent out
+ Server appears on the Steam Master List
+ Server can be joined 
+ A map can be played and restarted
+ Workshop items can be downloaded with `rd_enable_workshop_item`
+ Bonus: workshop updates on server are fixed

**What doesn't:**
- probably things that aren't tested and will show during gameplay

### Summarized
I think it's promising, and surprisingly, joining lobbies works from both ends. Probably, once this is tested further, other issues arise, but all these should be in the game code itself.


ywgATustcbbs:
I don't have write permission to repo. Therefore I created #1095 for further review.

mithrand0:
```
Network: IP 0.0.0.0, mode MP, dedicated Yes, ports 29001 SV / 27005 CL
Server is hibernating
Setting breakpad minidump AppID = 563560
SteamInternal_SetMinidumpSteamID:  Caching Steam ID:  76561197960265728 [API loaded no]
StartGameServer_Init resulted in success
maxplayers set to 8
---- Host_NewGame ----
Host_NewGame on map lobby
Unqualified = jacob_save_126_07_31_17_47_41.campaignsave
Creating new save with filename: save/jacob_save_126_07_31_17_47_41.campaignsave
New save created: save/jacob_save_126_07_31_17_47_41.campaignsave
Adding master server 208.64.200.65:27015
Adding master server 208.64.200.39:27011
Sending heartbeat to master server...
---- Host_Changelevel ----
```

mithrand0:
No crashes in a map playthrough. Server/client locally work.

Connection to Steam servers work after a bit of hazzle:
```
Connection to Steam servers successful.
   VAC secure mode is activated.
```

Server shows up in the Steam master server list as well:
```
   "servers": [
      {
        "addr": "<hidden>",
        "gmsindex": -1,
        "steamid": "90291952997103646",
        "gameport": 29001,
        "specport": 0,
        "name": "Gandalf's Secret Lab",
        "game_description": "Alien Swarm: Reactive Drop",
        "appid": 563560,
        "gamedir": "reactivedrop",
        "version": "7.1.4.6",
        "product": "reactivedrop",
        "region": 255,
 ```

File: src/game/shared/swarm/asw_player_experience.cpp [hunk 1]
```diff
@@ -449,7 +449,11 @@ void CASW_Player::RequestExperience()
 	{
 		if ( IsLocalPlayer( this ) )
 		{
+#if defined(STEAMAPPS_INTERFACE_VERSION008)
 			SteamUserStats()->RequestCurrentStats();
+#else
+			SteamUserStats()->RequestUserStats(SteamUser()->GetSteamID());
```
BenLubar: don't need to request current stats at all in the newer SDK; it happens automatically

src/game/client/swarm/c_rd_infection_deathmatch_stats.cpp
```diff
@@ -38,7 +38,11 @@ void CRD_Infection_Deathmatch_Stats::OnUpdate( C_RD_HUD_VScript *pHUD )
 
 	if ( !m_bStatsRequested && SteamUserStats() )
 	{
+#if defined(STEAMAPPS_INTERFACE_VERSION008)
```
BenLubar: don't need to request local stats at all on the newer SDK (which might actually already be the case)

shouldn't be using an unrelated define like this in either case

BenLubar:
can delete steam_api.json; we don't need it

src/game/shared/swarm/asw_player_experience.cpp
mithrand0: removed

mithrand0:
I think the most practical next step is to publish a `sdk165` test release with all the dll's, and begin testing everything from there. The concept works, and broader testing will tell us where adjustments are actually needed.

Any beautification or structural cleanup can follow at a later time.

src/game/server/gameinterface.cpp
```diff
@@ -619,6 +622,10 @@ bool CServerGameDLL::DLLInit( CreateInterfaceFn appSystemFactory,
 		CGlobalVars *pGlobals)
 {
 
+#ifndef CLIENT_DLL
```
BenLubar: this file isn't compiled with `CLIENT_DLL` - I think you should be asking the engine whether it's a dedicated server here instead

src/game/shared/swarm/rd_workshop.h
```diff
@@ -81,7 +81,9 @@ class CReactiveDropWorkshop : public CAutoGameSystem
 	void SetupThink();
 	bool DedicatedServerWorkshopSetup();
 	void EnableServerWorkshopItem( PublishedFileId_t id );
+	void OnPublishedFileDetails(RemoteStorageGetPublishedFileDetailsResult_t* pResult, bool bIOFailure);
 	bool m_bWorkshopSetupCompleted;
+	CUtlMap<PublishedFileId_t, int> s_SteamRemoteStorageChecked{DefLessFunc(PublishedFileId_t)};
```
BenLubar: this isn't static, so it shouldn't start with `s_`

src/game/client/swarm/c_rd_infection_deathmatch_stats.cpp
```diff
@@ -38,7 +38,6 @@ void CRD_Infection_Deathmatch_Stats::OnUpdate( C_RD_HUD_VScript *pHUD )
 
 	if ( !m_bStatsRequested && SteamUserStats() )
```
BenLubar: can probably remove this entire variable and the code that references it

src/game/client/cdll_client_int.cpp
```diff
@@ -2518,6 +2518,9 @@ void OnRenderEnd()
 
 void CHLClient::FrameStageNotify( ClientFrameStage_t curStage )
 {
+	// this needs to run on every frame
+	SteamAPI_RunCallbacks();
```
BenLubar: should probably only run callbacks once per frame (check for a specific value for `curStage`)

src/game/server/swarm/rd_gameserver.h
```diff
@@ -0,0 +1,14 @@
+#ifndef GAMESERVER_H
+#define GAMESERVER_H
+
+#ifdef _WIN32
+#pragma once
+#endif
+
+#include "cbase.h"
+#include "ws2tcpip.h"
```
BenLubar: include isn't used in this header, so it can be moved to the cpp file

src/game/shared/swarm/rd_workshop.h
mithrand0: renamed them

src/game/server/gameinterface.cpp
mithrand0: removed the check

src/game/client/cdll_client_int.cpp
mithrand0: only stages FRAME_START, FRAME_RENDER_START and FRAME_RENDER_END are always executed

since we need this in menu's etc as well, moved it to stage FRAME_START

src/game/server/swarm/rd_gameserver.h
mithrand0: moved them to the .cpp

PR #1092, opened 2026-08-17
Prevent debug assertions in controller focus and weapon details

ywgATustcbbs:
## Summary
- Avoid registering controller-focus panels while the client is outside an active game.
- Allow weapon detail tooltips to display challenge-defined reserve ammo that is not an exact multiple of the clip size.
- Handle non-positive clip sizes safely in the weapon detail UI.

### Controller focus outside a game
`CControllerFocus::AddToFocusList()` asserted `IsInGame()` before reaching its existing early return. The challenge-selection panel can be opened from the main menu before a listen server exists, so a Debug client hits the assertion even though focus registration should simply be skipped.

### Weapon detail reserve ammo
`CNB_Weapon_Detail::UpdateLabels()` asserted `nMaxBulletsPerGun % iClipValue == 0`. `MaxCarry()` is reserve ammunition, and challenge configurations may intentionally use a partial reserve clip (for example, 18 reserve rounds with an 8-round clip). Release builds already continued with integer-floor calculation, so exact divisibility is not a valid universal invariant.

## Fix
- Return from `AddToFocusList()` before the in-game assertion when `IsInGame()` is false.
- Do not enforce exact divisibility for reserve ammo; preserve the existing Release integer-floor display plus the loaded clip.
- Guard non-positive clip sizes, show `N/A`, and hide the stat bar.

## Scope
Client-side UI robustness only. No VScript, challenge behavior, weapon values, or ammo configuration changes.

## Reproduction
1. Debug client: open challenge selection from the main mission/task UI before creating a listen server.
2. Debug client: in briefing, inspect a weapon whose challenge-defined reserve ammo is not divisible by its clip size.

## Validation
- Win32 Debug client build passed.
- Win32 Release client build passed.
- `git diff --check` passed.

### DIFF
--- a/src/game/client/swarm/vgui/controller_focus.cpp
+++ b/src/game/client/swarm/vgui/controller_focus.cpp
@@ -77,11 +77,11 @@ void CControllerFocus::AddToFocusList( vgui::Panel *pPanel, bool bClickOnFocus, 
 	if ( !pPanel )
 		return;
 
+	if ( !IsInGame() )
+		return;
 	Assert( bModal || !m_iModalScope );
 
 	Assert( IsInGame() );
-	if ( !IsInGame() )
-		return;
 
 	FocusArea Focus;
 	Focus.hPanel = pPanel;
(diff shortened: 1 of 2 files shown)


ywgATustcbbs:
无标题2
无标题1

Debug, test, code review - human
Fix - fully AI generated

ywgATustcbbs:
This is a debug build only issue.

PR #1093, opened 2026-08-17
Add Traitors gameplay guide and bindable controls

ywgATustcbbs:
## Purpose / Summary

This change adds an in-game gameplay guide for the Traitors challenge. Every client can toggle it with the bindable `I` key while playing, dead, spectating, without a Marine, or after joining mid-game. The existing `O` skill-menu binding remains available and configurable.

## Implementation

- C++ adds only the bindable command/input routing and the local `rd_vgui_vscript` input opt-in.
- Server VScript creates the per-player menu entity at gameplay start and on `fullyjoined`, then only notifies the client when the menu should open or close.
- Client VScript owns the home page, second-level page navigation, close/reset behavior, and `ForceSync` handling.
- Localization note: The English and Simplified Chinese localization text was written by a human.
- Long guide sections are split into shorter localized tokens and concatenated for display; tips remain as individual entries.

## Safety / Lifecycle

- Menu entities are not created during briefing or mission selection, avoiding the debug UI assertion path that was previously exposed by early script startup.
- A local-player sentinel limits input control to the intended client.
- The guide does not depend on the client having a Marine, so dead and spectator clients remain eligible.

## Testing and Review

- `git diff --check` passes.
- Static review covered the lifecycle, local-player routing, synchronization, localization fallback, and menu reset behavior.
- The test addon was manually tested for opening and closing the guide, navigating its pages, displaying text, and coexisting with the `O` skill menu binding.
- Spectator, dead-player, and late-join paths were reviewed line-by-line at code level; they were not claimed as manual gameplay tests here.

Development disclosure: This change was produced with AI assistance under human-authored instructions. The resulting code was reviewed line-by-line by a human and manually tested by a human.

### DIFF
--- a/reactivedrop/content/traitors_challenge/scripts/vscripts/challenge_traitors.nut
+++ b/reactivedrop/content/traitors_challenge/scripts/vscripts/challenge_traitors.nut
@@ -142,6 +142,21 @@ g_bool_TraitorWinDelyFlag <- false; //Delay displaying winner msg
 
 g_ent_HudAndVGui <- []; // Entindex of all UI elements that have ever been created, used to release them at end of each mission. 存储所有挑战使用到的UI实体索引，用于游戏结束时释放资源
 
+// The gameplay guide is a player-owned VGUI screen.  Unlike the role skill
+// menus it is deliberately attached to the player entity instead of a marine:
+// spectators, dead players, and players who have not selected a marine yet can
+// all read it.  The client script resolves the guide tokens from the challenge
+// translation table; this entity only carries the client's language and the
+// open/closed state.
+g_str_GameplayGuideVGuiScript <- "challenge_traitors_client_gameplay_guide.nut";
+g_str_GameplayGuidePlayerField <- "strGameplayGuideVGuiName";
+g_str_GameplayGuideGenerationField <- "intGameplayGuideGeneration";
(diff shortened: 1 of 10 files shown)


ywgATustcbbs:
Updated the PR based on the review feedback:

- Replaced the duplicated keyboard-only bind control with the existing `CRD_VGUI_Bind`.
- Added normal Steam Input registrations:
  - `CustomFunction1` -> `scripted_user_func custom_function_1`
  - `CustomFunction2` -> `scripted_user_func custom_function_2`
- Generalized the command, variable, UI, and localization names so the bindings can be reused by other challenges.
- Added default bindings: O for Custom Function 1 and I for Custom Function 2.
- Updated the Traitors scripts to use the generic commands.
- Allowed both custom-function commands to bypass VGUI interception, while keeping the exception limited to these two exact commands.
- Made the player-owned `rd_vgui_vscript` path truly opt-in via its sentinel, preserving the original `m_hInteracter` behavior for other screens.
- Made custom-function binding comparisons case-insensitive.
- Did not modify the Steam Input manifest or controller configuration.

`git diff --check` passes, and the Release Win32 client target builds successfully.

Tested using keyboard and mouse. Joystick is not tested (I don't have one).

PR #1096, opened 2026-08-28
Modernize GitHub build workflow

anf3is:
- Merge jobs into a matrix for easier maintenance
- Add configurable workflow dispatch
- Update GitHub Actions to their latest versions

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -14,78 +14,38 @@ on:
 # https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions
 
 jobs:
-  release:
-    name: build release
+  build:
+    name: build ${{ matrix.configuration }}
     runs-on: windows-2025-vs2026
     timeout-minutes: 30
+    strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -8,84 +10,84 @@ on:
...
 # We can use it to build our project and do some automated builds and in future quality tests.
 # https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions
 
 jobs:
-  release:
-    name: build release
-    runs-on: windows-2025-vs2026
+  build:
+    name: build ${{ matrix.configuration }}
+    runs-on: windows-2025
```
mithrand0, line 47: Suggestion: keep the pin on 2026

If Microsoft ever releases 2028, then we should continue to use 2026 until we have made sure it compiles under 2028.

anf3is: As discussed on Discord, the latest `windows-2025` image already ships with vs2026, so there’s no need for an unconventional runner.

PR #1101, opened 2026-08-29
Add a faction for incapacitated marines

### LINKED ISSUE #1077: Freeze Grenade can unexpectedly freeze marines that have been knocked out and incapacitated.
To test, we can set these convars below:
```
rd_allow_revive 1
asw_skill_grenades_freeze_duration_base 100
```
Now kill a marine ( makes it incapacitated ), and then throw the marine a freeze grenade, and finally revive the marine.
And now you can see the marine is frozen.

anf3is:
Make a faction with asymmetric relationships.

- Same as alive marines towards others. Ensure that projectiles would work correctly even when they were launched before death but are still en route;
- Like from alive marines. Prevent friendly fire (and debuffs like freeze);
- Neutral from other even when they hate alive marines. Prevent crowding on dead bodies.

---

- Close #1077

- Depends on #1099


dragokas:
Hi dear developers.
I'm against "fixing" https://github.com/ReactiveDrop/reactivedrop_public_src/issues/1077

This is long-time behavior. Making grenade not to freezing incapacitated marine will cause significant change in gameplay balance: it will make the gameplay much more easier.

As a long-time player, I can see this change as very significant. The old behavior, where a player remains frozen after being revived, forces them to think, choose different tactics, and consider every move. If we "fix" this freezing, the player no longer has to think, no longer has to choose the right moment to revive, he can just "spam" with grenades, revive all players, and they will survive with almost 100% chance having a lot (too much) of time to defend.

anf3is:
Please don't spam in PR section when there is a relevant issue.

PR #1109, opened 2026-09-09
make spatial features of env_ambient_light work in asw

jhh8:
base player stays at spawn point...

### DIFF
--- a/src/game/client/c_spatialentity.cpp
+++ b/src/game/client/c_spatialentity.cpp
@@ -7,6 +7,8 @@
 #include "cbase.h"
 
 #include "c_spatialentity.h"
+#include "c_asw_player.h"
+#include "c_asw_inhabitable_npc.h"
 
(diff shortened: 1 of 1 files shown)


src/game/client/c_spatialentity.cpp
```diff
@@ -64,17 +66,17 @@ void C_SpatialEntity::ClientThink()
 		return;
 	}
 
-	CBaseEntity *pPlayer = C_BasePlayer::GetLocalPlayer( 0 );
-	if( !pPlayer )
+	C_ASW_Inhabitable_NPC* pViewNPC = C_ASW_Player::GetLocalASWPlayer()->GetViewNPC();
```
BenLubar, line 69: Do we have a static function that does this? I'm a little bit worried about a potential crash from the local player being null for some reason.

jhh8: my thoughts were it will probably crash all over different places in that case too

